PR #162, opened 2023-08-20
Created a breadcrumb component using fudis link components.

tommi-miettinen:
Tried to follow the contributing guidelines to my best ability.


joonashak:
Thanks for the PR. We will have someone from our team review the component code soon. In the meanwhile, however, please remove the changes to `.eslintrc.json` and `.prettierrc.json`. (It seems to me that your IDE is not correctly configured to either use prettier or respect it's default settings, you might be globally overriding them or something like that. `endOfLine` should be left to it's default value, otherwise we will have a huge diff created by prettier when it's properly used.)

joonashak (changes requested):
`.eslintrc.json` and `.prettierrc.json` should not be changed. Please see my comment in Conversation.

PR #195, opened 2023-10-04
Update workflows

joonashak:
- Run tests for main branch pushes and pull request events instead of just all pushes. This enables tests to be run after approval for external pull requests.
- Split linting and unit tests to separate jobs so they run independently.
- Run tests before publishing.
- Fixed CLA URL.


joonashak:
I have read the CLA Document and I hereby sign the CLA

PR #207, opened 2023-10-19
[CHORE]: Add current package version to Storybook and create release script

RiinaKuu:
When in local dev environment, you can see version from package.json

Screenshot 2023-10-19 at 13 55 44

----------------------------------
When in deployed Storybook from the branch, you can see version with the branch name. This is added to avoid confusion when sharing links to the branch's deployed Storybook.

Screenshot 2023-10-20 at 16 52 12


ngx-fudis/.storybook/getVersion.js
```diff
@@ -11,14 +12,14 @@ const defaultVersion = {
...
-	if (!isValidSemVer(tagOrBranchName)) {
+	if (!isValidSemVer(pkg.version)) {
 		return {
 			...defaultVersion,
 			postfix: tagOrBranchName,
 		};
 	}
 
-	return parseSemVer(tagOrBranchName);
+	return parseSemVer(pkg.version);
```
joonashak: If `pkg.version` is returned, other logic in this file is stale, so should be removed.

However, this change will mean that documentation built from PR branches will be labelled with the latest version which might or might not be what we want. I should note also that this change is not necessary as the build system has control over the label with the current setup (which makes configuring the workflows nice).

joonashak (dismissed):

PR #216, opened 2023-10-27
[Link]: Underline appearance enhancement

RiinaKuu:
Same link appearance as in Sisu (relates to https://jira.funidata.fi/browse/OTM-29921)
Link should always have solid underline, no hover changes.


ngx-fudis/projects/ngx-fudis/src/lib/components/link/link.component.ts
```diff
@@ -32,7 +32,7 @@ import { FudisFocusService } from '../../services/focus/focus.service';
  * <fudis-link
  *     [href]="https://www.example.com"
  *     [isExternalLink]="true"
- * 	color="default">
+ * 	[color]="'default'">
  * </fudis-link>
```
videoeero, line 36: I think we should aim to remove these *.ts examples, and focus on keeping Storybook documentation updated.

RiinaKuu: Removed those examples from typescript and updated documentation by adding readme.mdx

PR #256, opened 2024-01-30
Visual regression tests

joonashak:


joonashak:
Cypress goes into :roskis: in favor of Playwright #271

PR #277, opened 2024-02-14
[Id, Translation and LanguageBadgeGroup Services]: Checking for 1.0

RiinaKuu:
https://funidata.atlassian.net/browse/DS-240

Added TODO comments and filled up documentation.


ngx-fudis/projects/ngx-fudis/src/lib/services/translation/readme.stories.mdx
```diff
@@ -0,0 +1,9 @@
+import { Meta } from "@storybook/blocks";
+
+<Meta title="Services/Translation" />
+
+# Translation Service
+
+{/* prettier-ignore */}
+{/* TODO: Write documentation */}
+_Coming soon_
```
MayaMarjut, line 9: Onko tää siis tarkoituksella jätetty muotoon readme.stories ettei tämä näkyisi?

RiinaKuu: This is intentional. How I pictured this, is that we want the users to know that there will be documentation for this. Translation service is something that is already exposed and it _has_ to be exposed even though the documentation or tests are lacking.

PR #278, opened 2024-02-14
[Storybook, Deps]: Hotfix for restoring argtypes

videoeero:
- updated deps
- removed few storybook addons
- adjusted compodoc configs


joonashak (approved):
Very nice, I like it

PR #283, opened 2024-02-15
[Datepicker & DateRange]: Checking for 1.0

RiinaKuu:
https://funidata.atlassian.net/browse/DS-240

BREAKING:
- Hid `Datepicker` from public API and from Storybook
- Hid `DateRange` from public API and from Storybook

Important date common/utility functions were missing tests.
Also deleted unused Datepicker Custom Header Component which was there just in case we needed it earlier when Datepicker was in development.


ngx-fudis/projects/ngx-fudis/src/lib/components/form/date/date-range/readme.hidden
```diff
@@ -1,7 +1,7 @@
 import { ArgTypes, Meta, Canvas } from "@storybook/blocks";
 import { linkTo } from "@storybook/addon-links";
-import { DateRangeComponent } from "./date-range.component.ts";
-import * as DateRangeStories from "./date-range.stories.ts";
+import { DateRangeComponent } from "./date-range.component.js";
```
MayaMarjut, line 3: Why?

PR #322, opened 2024-03-14
[DropdownItemBase]: Fix for unwanted dropdown close when multiselect is a child component of fieldset

videoeero:
Noticed a bug, when Multiselect was used as child of Fieldset and added some duck tape fix to it. The function _focusedOutFromComponent is now quite a spaghetti and should be rethought before releasing Select family for >1.0


videoeero:
Closing for now, linked this PR in other related ticket for reference

PR #365, opened 2024-05-20
[Datepicker, DateAdapter]: Expose Datepicker component to public API, tests

RiinaKuu:
https://funidata.atlassian.net/browse/DS-188

- Expose Datepicker to public API
- Unit and Playwright test for Datepicker
- Unit tests for DateAdaprter


videoeero (approved):
Approving now as date-range related stuff was post-poned to another ticket

PR #392, opened 2024-06-19
[DropdownMenu]: Introduce new Dropdown Menu component

RiinaKuu:
Jira-ticket: https://funidata.atlassian.net/browse/DS-260

- Refactor the old DropdownMenuComponent
- Expose the component to public API

Tests will be written in another PR, https://github.com/funidata/fudis/pull/405

**Known issues atm:**
- [Safari does not focus buttons when clicking by design](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#clicking_and_focus). First tried to give focus manually but that broke the whole event flow. Since there is no focus from the start, outside click does not work at all in Safari.


ngx-fudis/projects/ngx-fudis/src/lib/components/button/button.component.html
```diff
@@ -1,24 +1,28 @@
 <ng-container *ngIf="tooltip">
   <button
     #buttonElement
+    fudisTooltip
+    [class]="_classList | async"
     [type]="type"
     (blur)="_handleButtonBlur($event)"
     (focus)="_handleButtonFocus($event)"
+    (click)="!disabled && buttonClick($event)"
```
MayaMarjut, line 9: !disabled here doesn't have any effect

ngx-fudis/projects/ngx-fudis/src/lib/components/button/button.component.html
```diff
@@ -1,24 +1,28 @@
...
+    (keydown)="_handleMenuButtonKeyDown($event)"
     [attr.aria-disabled]="disabled"
-    [attr.aria-label]="_ariaLabel"
+    [attr.aria-label]="_ariaLabel | async"
     [attr.id]="_id"
-    (click)="!disabled && buttonClick($event)"
-    [ngClass]="_classList"
-    fudisTooltip
+    [attr.aria-expanded]="!asMenuButton ? null : dropdownOpen"
```
MayaMarjut, line 14: Expanded and collapsed state is not transfered properly. Now aria expanded get's state as [object object]. Seems like this state is not properly tied to handleclick function.

MayaMarjut: Currently when focusing on closed menu-button, screen readers reads it as expanded, when should be collapsed.

ngx-fudis/projects/ngx-fudis/src/lib/components/button/button.component.html
```diff
@@ -31,20 +35,24 @@
...
     (focus)="_handleButtonFocus($event)"
+    (keydown)="_handleMenuButtonKeyDown($event)"
     [attr.aria-disabled]="disabled"
-    [attr.aria-label]="_ariaLabel"
+    [attr.aria-label]="_ariaLabel | async"
     [attr.id]="_id"
-    (click)="!disabled && buttonClick($event)"
-    [ngClass]="_classList"
+    [attr.aria-expanded]="!asMenuButton ? null : dropdownOpen"
+    [attr.aria-controls]="asMenuButton && dropdownMenuId ? dropdownMenuId : null"
```
MayaMarjut, line 48: Consider adding aria-haspopup over here

ngx-fudis/projects/ngx-fudis/src/lib/components/button/button.component.html
```diff
@@ -31,20 +35,24 @@
 <ng-container *ngIf="!tooltip">
   <button
```
MayaMarjut, line 36: Currently tab-key doesn't work at all on Safari. Try adding tabindex=0 here, which makes button accessible with Safari browser

ngx-fudis/projects/ngx-fudis/src/lib/components/dropdown-menu/dropdown-menu-item/dropdown-menu-item.component.html
```diff
@@ -1,54 +1,22 @@
 <button
-  *ngIf="!_parentComponent.multiselect"
   #dropdownItem
-  class="fudis-dropdown-menu-item fudis-dropdown-menu-item__focusable"
-  [attr.role]="'menuitem'"
+  class="fudis-dropdown-menu-item"
+  [class.fudis-dropdown-menu-item--disabled]="disabled || null"
+  [class.fudis-dropdown-menu-item__focusable]="!disabled"
   [attr.id]="_id"
```
MayaMarjut, line 6: As a usability improvement I would add  tabindex="-1" here, which would make menu items unfocusable by tab key, while tab key would be preserved for moving out of menu. This way user doesn't have to tab through all menu items in order to exit while only tabing forward.

ngx-fudis/projects/ngx-fudis/src/lib/components/button/button.component.html
RiinaKuu: If the button is disabled we don't want click to emit any event. It is necessary attribute.

ngx-fudis/projects/ngx-fudis/src/lib/components/button/button.component.html
RiinaKuu: `dropdownOpen` observable was lacking async pipe in the template, fixed.

ngx-fudis/projects/ngx-fudis/src/lib/components/button/button.component.html
RiinaKuu: Good addition, added.

ngx-fudis/projects/ngx-fudis/src/lib/components/button/button.component.html
RiinaKuu: Tab key is working fine for me on Safari as well.

ngx-fudis/projects/ngx-fudis/src/lib/components/dropdown-menu/dropdown-menu-item/dropdown-menu-item.component.html
RiinaKuu: Didn't know this was a good practice. Good to know, added it.

MayaMarjut (dismissed):

MayaMarjut (dismissed):

MayaMarjut (approved):

PR #408, opened 2024-07-11
[Dependencies]: Update

RiinaKuu:
Jira-tikcet: https://funidata.atlassian.net/browse/DS-321

- Added `@types/react` library inside overrides since there were clashing versions in Storybook (`npm list @types/react`, 18.3.3 vs 16.14.60)
- Updated node version in `.nvmrc` from `18.18.2` --> `18.19.1`


ngx-fudis/package.json
```diff
@@ -23,53 +23,56 @@
...
-    "storybook": "^7.6.19",
-    "stylelint": "^16.6.0",
+    "storybook": "^7.6.20",
+    "stylelint": "^16.6.1",
     "stylelint-config-concentric-order": "^5.2.0",
     "stylelint-config-sass-guidelines": "^11.1.0",
     "stylelint-config-standard-scss": "^13.1.0",
     "ts-node": "^10.9.2",
-    "typescript": "^5.2.2"
+    "typescript": "^5.4.5"
```
joonashak, line 73: I would also update `tslib` and `typescript` in `devDependencies` of `ngx-fudis/projects/ngx-fudis/package.json` to match the versions specified here. Probably no biggie, but in edge cases might avoid some weird behavior while developing. And anyway [`devDependencies` changes are not considered breaking changes](https://softwareengineering.stackexchange.com/questions/440018/semantic-versioning-on-a-dev-dependencies-change) so should be trivial to keep them in sync with this file.

joonashak (approved):

PR #440, opened 2024-09-18
[Grid, Section, DL]: Remove marginTop and marginBottom input properties from Grid, change classes input to type from array to string

xanxan:
https://funidata.atlassian.net/browse/DS-368

BREAKING CHANGE GRID:
- Remove marginTop and marginBottom input properties
- Change classes type from string[] to string
- Remove references from tests, documentation and stories
- Replace usages with fudis spacing classes
- Refactor Grid directive to inherit classes given through the class property, so the spacing classes can be applied

BREAKING CHANGE SECTION & DL
- Change classes type from string[] to string

BEAKING CHANGE SPACING TOKENS
- Renamed form field related "spacing" SCSS tokens to better describe their purpose.


ngx-fudis/projects/ngx-fudis/src/lib/components/dialog/dialog.stories.ts
```diff
@@ -180,7 +180,7 @@ class DialogWithFormComponent {
             </fudis-body-text>
           </fudis-grid>
           <hr />
-          <fudis-grid [columns]="{ xs: 1, sm: 2, md: 3 }" [marginTop]="'sm'" [marginBottom]="'sm'">
+          <fudis-grid [columns]="{ xs: 1, sm: 2, md: 3 }" [classes]="['fudis-mt-sm fudis-mb-sm']">
```
RiinaKuu, line 183: Change above `<hr />` element to `<fudis-hr />`.
Grid columns' xs is always one so let's remove `xs: 1 `part from the JSON object.

ngx-fudis/projects/ngx-fudis/src/lib/components/dialog/dialog.stories.ts
```diff
@@ -154,7 +154,7 @@ class DialogWithFormComponent {
           >Dialog with fudis-grid and scrollable content</fudis-heading
         >
         <fudis-dialog-content>
-          <fudis-grid [marginTop]="'md'" [marginBottom]="'md'">
+          <fudis-grid [classes]="['fudis-mt-md fudis-mb-md']">
```
RiinaKuu, line 157: Seems like this margin-top does not affect at all, `mat-dialog__content` class is overriding it.

ngx-fudis/projects/ngx-fudis/src/lib/components/description-list/description-list.stories.ts
```diff
@@ -117,11 +117,11 @@ const NestedDescriptionListsTemplate: StoryFn<DescriptionListComponent> = (
     <fudis-heading [level]="2" [variant]="'md'"
       >Nested Description Lists with Indiana Jones Movies</fudis-heading
     >
-    <fudis-dl class="fudis-mt-sm" [disableGrid]="disableGrid">
+    <fudis-dl class="fudis-mt-sm" [disableGrid]="disableGrid" [classes]="['fudis-mb-none']">
       <fudis-dl-item>
         <fudis-dt [contentText]="'Raiders of the Lost Ark'"></fudis-dt>
         <fudis-dd>
-          <fudis-dl [variant]="'compact'">
+          <fudis-dl [variant]="'compact'" [classes]="['fudis-mt-none fudis-mb-none']">
```
RiinaKuu, line 124: In general I think this is a bad pattern.
Browser adds default styles (margins) to native `<dl>` element which we used to override with Grid's margin-top and margin-bottom values. I think we still want to override those values coming from the browser, so some king of logic for that should be added where `[classes]` input's values will then override the initial none values.

RiinaKuu (changes requested):

videoeero (approved):

RiinaKuu (approved):

PR #444, opened 2024-09-24
[Dialog]: New responsive screen sizes and renamed spacing token input widths

MayaMarjut:
https://funidata.atlassian.net/browse/DS-370

Dialog's old size variant 'initial' was removed and was replaced by a new small mobile screen size 'xs'. Unit tests and documentation was updated.

All screen sizes were updated to align with a new design. 

Changes were made to form grid with xxl within dialog. The default xxl grid width didn't span the whole width of a dialog size xxl. Now grid size xxl does an exception within form dialog class. 

Component and input specific  width tokens were renamed in spacing tokens file. These changes included renaming radio button, radio-button indicator and checkbox width tokens. Dialog specific width tokens were created and stored in dialog scss file. 

New story was created for VR testing purposes for all dialog sizes.


ngx-fudis/projects/ngx-fudis/src/lib/components/dialog/dialog.component.scss
```diff
@@ -1,65 +1,79 @@
...
+}
 
 .fudis-dialog {
   box-sizing: content-box;
   position: relative;
-  margin: spacing.$spacing-md spacing.$spacing-sm;
-  max-width: calc(100vw - (2 * spacing.$spacing-sm));
+  padding: spacing.$spacing-md spacing.$spacing-xl;
+  /* stylelint-disable-next-line unit-disallowed-list */
+  max-width: calc(100vw - 10rem);
```
videoeero, line 17: We cannot use rem values without also including --fudis-rem-multiplier variable here. Without it, this 10rem will result in 160px by default and 100px in e.g. SISU.

ngx-fudis/projects/ngx-fudis/src/lib/components/dialog/dialog.component.scss
```diff
@@ -1,65 +1,79 @@
...
+      @include breakpoints.breakpoint('sm') {
+        width: calc(spacing.$spacing-dialog-xl - (2 * spacing.$spacing-xl));
+      }
 
-  @include breakpoints.breakpoint('md') {
-    margin: spacing.$spacing-md spacing.$spacing-xl;
-    max-width: calc(100% - (2 * spacing.$spacing-xl));
+      @include breakpoints.breakpoint('lg') {
+        /* stylelint-disable-next-line unit-disallowed-list */
+        max-width: calc(100vw - 15rem);
```
videoeero, line 75: Same here, needs --fudis-rem-multiplier for calculation.

ngx-fudis/projects/ngx-fudis/src/lib/components/dialog/dialog.component.scss
```diff
@@ -1,65 +1,79 @@
...
+
+      @include breakpoints.breakpoint('sm') {
+        width: calc(spacing.$spacing-dialog-lg - (2 * spacing.$spacing-xl));
+      }
 
-      @include breakpoints.breakpoint('md') {
-        width: calc(gridTokens.$grid-width-lg + 2 * spacing.$spacing-xl);
+      @include breakpoints.breakpoint('lg') {
+        /* stylelint-disable-next-line unit-disallowed-list */
+        max-width: calc(100vw - 15rem);
```
videoeero, line 62: Also here --fudis-rem-multiplier calculation is needed.

ngx-fudis/projects/ngx-fudis/src/lib/components/dialog/ngMaterial-theme.scss
```diff
@@ -91,6 +95,16 @@
...
           }
         }
+
+        & fudis-form {
+          & .fudis-grid__xxl {
+            @include breakpoints.breakpoint('sm') {
+              @include grid.grid-max-width('xxl');
+
+              gap: spacing.$spacing-sm;
+            }
```
videoeero, line 105: I'd like to see and understand reasoning behind this! I understand that for the "first level" grid this is okay, but what if Form has nested grids? This would affect those as well. Is that intended?

ngx-fudis/projects/ngx-fudis/src/lib/foundations/spacing/tokens.scss
```diff
@@ -31,3 +30,10 @@ $spacing-input-lg: calc(23rem / var(--fudis-rem-multiplier));
 $spacing-input-checkbox: calc(1.25rem / var(--fudis-rem-multiplier));
 $spacing-input-radio: calc(1.25rem / var(--fudis-rem-multiplier));
 $spacing-input-radio-indicator: calc(0.75rem / var(--fudis-rem-multiplier));
+
+// Dialog size widths
+$spacing-dialog-xs: calc(30rem / var(--fudis-rem-multiplier));
+$spacing-dialog-sm: calc(44rem / var(--fudis-rem-multiplier));
+$spacing-dialog-md: calc(55rem / var(--fudis-rem-multiplier));
+$spacing-dialog-lg: calc(70rem / var(--fudis-rem-multiplier));
+$spacing-dialog-xl: calc(90rem / var(--fudis-rem-multiplier));
```
videoeero, line 39: Not necessarily something that needs to be fixed, but I think we should discuss with the team if these naming conventions make sense. As usually term "spacing" refers to margins and paddings between elements, but now these are used for sizes as well. I think they can live in this file, but as 3.0 is already a breaking change, I think we should rename these as well to go with the flow with 3.0.

ngx-fudis/projects/ngx-fudis/src/lib/components/dialog/ngMaterial-theme.scss
MayaMarjut: The reason behind this is that we want form action buttons to always flow to dialog's right corner when dialog is enlargened. Without this style, form within a dialog would have buttons hanging in a middle of large dialogs and this is not designer's intention. The default gap size also led to an unnecessary increase in dialog height.

ngx-fudis/projects/ngx-fudis/src/lib/foundations/spacing/tokens.scss
videoeero: Discussed and renamed/refactored these

ngx-fudis/projects/ngx-fudis/src/lib/components/dialog/ngMaterial-theme.scss
videoeero: Discussed and renamed/refactored these

videoeero (approved):

PR #445, opened 2024-09-24
[RadioButton, Checkbox]: Unify output namings, refactor radio button and checkbox blur and change events

xanxan:
https://funidata.atlassian.net/browse/DS-372

BREAKING CHANGE
- Unify output names to use handle prefix
- Remove radioButtonBlur output from RadioButton component, call parent onBlur instead
- Add handleChange output for RadioButton component
- On Checkbox Component, add input blur to call also parent onBlur


ngx-fudis/projects/ngx-fudis/src/lib/components/form/select/common/autocomplete/autocomplete.component.ts
```diff
@@ -91,47 +91,47 @@ export class SelectAutocompleteComponent implements OnChanges, OnInit {
   /**
    * Output event for input field blur
    */
-  @Output() triggerBlur = new EventEmitter<FocusEvent>();
+  @Output() handleBlur = new EventEmitter<FocusEvent>();
```
videoeero, line 95: We should discuss with the team, if this name change is wanted. Autocomplete component is not a public component, so it does not have to have same API naming conventions as public components. Actually prefix `trigger` was intentionally used here to differentiate it from public `handle` outputs.

videoeero (approved):

PR #448, opened 2024-09-28
[Form Common API]: Remove disabled & invalidState properties

videoeero:
Jira-ticket: https://funidata.atlassian.net/browse/DS-296

BREAKING CHANGE:
- Affecting all form components (text-input, datepicker etc.) `disabled` and `invalidState` properties are removed. Disabling should be done straight from FormControl. InvalidState was previously used internally, but is no longer needed and form field's invalid state should come from FormControl as well.

IMPROVEMENT LocalizedTextGroup
- If whole FormGroup is disabled (formGroup.disable()), it will disable the associated Select Component as well


MayaMarjut (approved):

ngx-fudis/projects/ngx-fudis/src/lib/components/form/localized-text-group/localized-text-group.component.html
```diff
@@ -80,6 +80,7 @@
   </div>
   <fudis-select
     *ngIf="_languageLabel | async as label"
+    #selectRef
     class="fudis-localized-text-group__select"
     [label]="label"
     [selectionClearButton]="false"
```
MayaMarjut, line 86: Eventhouth now select is disabled when all controls have disabled state. Still user is able to click the select chevron . Nothing is happening but the chevron rotates.

MayaMarjut (approved):

PR #460, opened 2024-10-09
[Angular Material]: Updating to Material 3

MayaMarjut:
https://funidata.atlassian.net/browse/DS-373

Breaking Change: Updating Angular peer dependencies as follows:
    "@angular/cdk": "^18.2.8",
    "@angular/common": "^18.2.7",
    "@angular/core": "^18.2.7",
    "@angular/material": "^18.2.8"

Updated [Angular Material 3](https://material.angular.io/) to Fudis. Removed old ngMaterial m2 theme-, palette and typography config files. New m3 theme was created and a new palette was generated. Some palette's hue inconsistency is a work around to match dialog background and datepicker period button default styles with the previous design. Fudis Angular Material component's styles were refactored and specially datepicker design was changed.

Also noted that previous dialog used font-size 1.125rem as default base style (set in palette). None of component's inherited this size, but removing this default caused layout to span. Decided to stick with this default and added font-size, line-height and  letter-spacing override default to dialog.


ngx-fudis/projects/ngx-fudis/src/lib/components/dialog/ngMaterial-theme.scss
```diff
@@ -113,9 +121,10 @@
   @include colorMixins.bg-color('white');
 
   &.cdk-overlay-backdrop.cdk-overlay-backdrop-showing {
+    top: -50vh;
```
RiinaKuu, line 124: Not sure what this does/fixes, if we have increased the `height` from `300vh` to `350vh` and then added `top: -50vh`?

ngx-fudis/projects/ngx-fudis/src/lib/components/form/date/datepicker/ngMaterial-theme.scss
RiinaKuu, line 1: `.mat-datepicker-close-button` seems to get wrong color, should be primary.

Screenshot 2024-10-10 at 14 56 16

ngx-fudis/projects/ngx-fudis/src/lib/theme/palette.scss
```diff
@@ -1,29 +1,117 @@
...
+// Note: Color palettes are generated from primary: #1076db, secondary: #d4d4d4
+$_palettes: (
+  primary: (
+    0: colors.$color-black,
+    10: colors.$color-primary-extra-dark,
+    20: colors.$color-primary-extra-dark,
+    25: colors.$color-primary-dark,
+    30: colors.$color-primary-dark,
+    35: colors.$color-primary-dark,
+    40: colors.$color-primary-light,
```
RiinaKuu, line 17: Is this intentional that hue 40 is `primary-light` and not `primary`?

ngx-fudis/projects/ngx-fudis/src/lib/theme/palette.scss
```diff
@@ -1,29 +1,117 @@
+/* stylelint-disable  @stylistic/no-missing-end-of-source-newline */
+@use 'sass:map';
 @use '@angular/material' as mat;
```
RiinaKuu, line 3: I would guess that we don't need this import here.

ngx-fudis/projects/ngx-fudis/src/lib/theme/theme-m3.scss
```diff
@@ -0,0 +1,23 @@
+/* stylelint-disable @stylistic/no-missing-end-of-source-newline */
```
RiinaKuu, line 1: This stylelint-disable is redundant.

RiinaKuu (changes requested):

ngx-fudis/projects/ngx-fudis/src/lib/components/dialog/ngMaterial-theme.scss
MayaMarjut: This is a hack that centres the viewport, so that the material dialog's background overlay won't overflow on Mobile-Chrome-Big-Landscape view

RiinaKuu (approved):

PR #461, opened 2024-10-09
[Footer]: Fix focus area

RiinaKuu:
https://funidata.atlassian.net/browse/DS-354

Links had unnecessary big focus area inside Footer
--> removed `min-height` from the links and adjusted margins


videoeero (approved):
LGTM

PR #467, opened 2024-10-17
[AlertGroup]: Missing documentation about template usage

RiinaKuu:

- Added documentation for using `<fudis-alert-group />` in the application template.
- Moved AlertGroup to the bottom element of Dialog Component

### DIFF
--- a/ngx-fudis/projects/ngx-fudis/src/lib/components/alert/alert.mdx
+++ b/ngx-fudis/projects/ngx-fudis/src/lib/components/alert/alert.mdx
@@ -11,7 +11,7 @@ Alert Group Component displays list of toaster-like Alert Components with four v
 
 ## Usage Guidelines
 
-First, add `<fudis-alert-group />` HTML tag in the application template, preferably somewhere up in the application's DOM, e.g. `app.component.html`.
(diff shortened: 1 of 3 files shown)


ngx-fudis/projects/ngx-fudis/src/lib/components/alert/alert.mdx
```diff
@@ -9,6 +9,12 @@ import { AlertGroupComponent } from "./alert-group/alert-group.component.ts";
 
 Alert Group Component displays list of toaster-like Alert Components with four variants: `success`, `info`, `warning` and `danger.` These variants are the same ones as with [Notification Component](/docs/components-notification--documentation).
 
+## Usage Guidelines
+
+First, add `<fudis-alert-group />` HTML tag in the application template, preferably somewhere up in the application's DOM, e.g. `app.component.html`.
```
videoeero, line 14: For the DOM structure we should probably instruct to put component as the "bottom element" of app.component.html etc. as that will make it more likely, that it will appear as the top element, even top of Navigation. This probably should be fixed in our examples and in our Dialog component as well, that Alert Group is last and not first element in DOM.

videoeero (approved):

PR #468, opened 2024-10-21
[Form Components]: Add "disabled" Input

RiinaKuu:

We came to the realisation that customer application needs to be able to disable form components as an input property. In few cases form had complex validation logic and disabling just the bound FormControl was cumbersome.

- Added "disabled" Input to Form Common API
- Added necessary props to all form component templates to enable new Input use

Nevertheless, the preferred way to set disabled status is through FormControl.


ngx-fudis/projects/ngx-fudis/src/lib/components/form/date/datepicker/datepicker.component.html
```diff
@@ -21,8 +21,8 @@
       [min]="_minDate"
       [max]="_maxDate"
       [placeholder]="_placeholderString | async"
-      [attr.disabled]="control.disabled || null"
-      [attr.aria-disabled]="control.disabled || null"
+      [attr.tabindex]="disabled ? -1 : 0"
```
videoeero, line 24: Should this be `[attr.tabindex]="disabled ? -1 : null"` ?

ngx-fudis/projects/ngx-fudis/src/lib/components/form/localized-text-group/localized-text-group.component.html
```diff
@@ -30,12 +30,15 @@
         (blur)="handleBlur.emit($event)"
         (focus)="onFocus($event)"
         [id]="id"
+        [attr.tabindex]="disabled ? -1 : 0"
```
videoeero, line 33: Should this be [attr.tabindex]="disabled ? -1 : null" ?

ngx-fudis/projects/ngx-fudis/src/lib/components/form/localized-text-group/localized-text-group.component.html
```diff
@@ -57,10 +60,11 @@
...
```
videoeero, line 63: ...and so on

ngx-fudis/projects/ngx-fudis/src/lib/components/form/date/datepicker/datepicker.component.html
```diff
@@ -32,13 +32,13 @@
       class="fudis-datepicker__toggle"
       [id]="id + '-calendar-icon-toggle'"
       [for]="picker"
-      [disabled]="control.disabled"
+      [disabled]="control.disabled || disabled"
```
videoeero, line 35: Is this necessary, as FormControl is binded to the input element, so this probably should be removed?

PR #470, opened 2024-10-22
[Select, Multiselect]: New input properties 'autocompleteFilter' and 'autocompleteNoResultsText' with Storybook examples

videoeero:
Jira ticket: https://funidata.atlassian.net/browse/DS-344

Fudis application has a need that Select Autocomplete variants selectable options are not loaded instantly in HTML, but only after user input has been evaluated and filtered matching options from the backend. This PR adds support for this, as new prop can disable default autocomplete filtering logic and give application more freedom on this. Also "No results found" default text can be set by application to for example "Searching results..." while backend search is in progress.

Also updated tests, docs and added Storybook examples.


RiinaKuu (commented):
When selecting one item from the list, blurring and focusing to the input again, dropdown opens but it shows the first option of the whole data and not the selected one:

Screenshot 2024-10-25 at 11 52 12

RiinaKuu (approved):

PR #471, opened 2024-10-22
[Datepicker, Tooltip]: Angular Material based components not reflecting all styles correctly in customer app

RiinaKuu:

Certain typography styles did not reflect correctly in customer (Sisu) application side. This might be due to Material 3 update where most of the typography configs were removed and M3 did not have clear instructions how to set up custom typography for individual components.

To fix the issue, some typography styles are overwritten in Datepicker and Tooltip components'  respective `ngMaterial-theme.scss` files. These changes should not effect the components in any way in Fudis.


videoeero (dismissed):
LGTM

videoeero (approved):

PR #472, opened 2024-10-22
[Typography]: Unify size syntax

RiinaKuu:

Rename and unify typography tokens from single char to double char, e.g. `$heading-l-font-size` --> `$heading-lg-font-size`


videoeero (approved):
LGTM

PR #495, opened 2024-11-28
[Expandable]: Remove default level value

RiinaKuu:
https://funidata.atlassian.net/browse/DS-396

Expandable should be part of the semantic order of the page. Initially Expandable's title had a default `aria-level` value of 2 but since we want to make sure that developers take into consideration of the semantic order, the default value was removed and `level` was set to be required Input property.

Updated docs and added few unit tests.


videoeero (approved):
LGTM

PR #496, opened 2024-11-28
[Section, Form]: Remove default level and titleVariant values, make title and level as required properties

RiinaKuu:
https://funidata.atlassian.net/browse/DS-397

1. Section is part of the semantic order of the page. Initially Section title had a default `aria-level` value of 2 and variant of 'lg', but since we want to make sure that developers take into consideration the semantic order, following changes have been made:
  - Set `level` as required property
  - Removed `level` and `titleVariant` default values 
  - Adjust `titleVariant` logic to match the corresponding `level` which can be overriden

2. Same kind of changes in Form Component as well
  - Set `title` and `level` as required properties
  - Removed `titleVariant` default value and adjust its logic to match the corresponding `level` which can be overriden

3. Updated docs and created Typography util function which is used in Heading and Section components.


videoeero (approved):
LGTM

PR #502, opened 2024-12-03
[Error Message]: Change component to directive

videoeero:
Jira: https://funidata.atlassian.net/browse/DS-404

Error Message Component doesn't have any HTML template, so it was unnecessary to keep it as Component. This PR changes it to Directive.

Breaking change migration:
- If application has imported `ErrorMessageComponent` from Fudis, it should now import `ErrorMessageDirective` instead


RiinaKuu (approved):
Discussed about and decided to keep this directive under components folder since it is used component-like with HTML selector.

PR #501, opened 2024-12-03
[Checkbox Group, Checkbox]: Adjust Checkbox Group's FormGroup and Checkbox's FormControl type

videoeero:
Jira ticket: https://funidata.atlassian.net/browse/DS-410

Application developers encountered Typescript typing error with Checkbox Group. After discussion this PR's adjustments were made to address the issue.

Breaking changes:
- Previously FormControl of Checkbox could be type of `boolean | null | undefined`. After this PR it can be `boolean | null` without undefined.
- CheckboxGroup component now requires type parameter of FormGroup's type
  - E.g. before: `element: CheckboxGroup` --> `element: CheckboxGroup<FudisCheckboxGroupFormGroup>` or `element: CheckboxGroup<MyType>`

Slack: https://sisudev.slack.com/archives/C03G46K6LG1/p1733216094386739

PR #504, opened 2024-12-04
[Checkbox Group]: Change FormGroup typing

videoeero:


videoeero:
Discussed with team and app developers, that let's use [ver1](https://github.com/funidata/fudis/pull/501) as base solution. So closing this for now!

PR #510, opened 2024-12-12
[Dialog]: Detect if dialog is scrollable

MayaMarjut:
https://funidata.atlassian.net/browse/DS-379

Added new componentType reference to domUtilities in order to separate which utility functions are needed. Created signal dialogScrollable to determine if dialog content is scrollable. The new signal is used for adding dynamically 'fudis-dialog--scrollable' -class to dialog content. This class is used for adding extra padding-right for the possible dialog content scrollbar.


ngx-fudis/projects/ngx-fudis/src/lib/components/dialog/ngMaterial-theme.scss
```diff
@@ -67,6 +67,7 @@
       .mat-mdc-dialog-content {
         @include utilities.box-reset;
 
+        padding-right: spacing.$spacing-md; /* Padding for a content scrollbar */
```
videoeero, line 70: My first thought was this solution as well, but did you ask designers if adding padding and make the container narrower is okay? Other solution could be "expanding" the content so that scrollbar would appear "outside" of container but the width of content itself would stay the same. Also is md/24px padding necessary? Wouldn't smaller amount be enough? I think this is an edge case where non-8px amount is acceptable. Eg. 18px, where 1rem is the usual scrollbar width and 2px just to give some breathing space if needed. Also did you consider if we could dynamically add this padding only if content requires scrolling? We previously had a working function to check scrollability which was used in other, now redundant, issue.

RiinaKuu (approved):

PR #518, opened 2024-12-18
[Link Directive]: Changed default size

MayaMarjut:
https://funidata.atlassian.net/browse/DS-379

Previous default size 'inherit' caused problems in client services. Changed default size to 'md', since it's also the default size of fudis-body-text component. Also now, if link is wrapped inside of fudis-body-text component it will automatically inherit the font size of it's parent. Fixed bug in dialog material scss and fixed type issue in sandbox. Updated VR images.


ngx-fudis/projects/ngx-fudis/src/lib/directives/link/link.directive.stories.ts
```diff
@@ -86,67 +86,37 @@ export const AllCombinations: StoryFn = (args) => ({
...
-          [size]="'inherit'"
-          [title]="'external link of size inherit'"
-        ></a>
+        <a fudisLink href="" [external]="true" [size]="'inherit'" [title]="'external link'"></a>
         as its child elements.</fudis-body-text
       >
       <fudis-body-text [variant]="'lg-light'"
         >This Body Text lg-light has
-        <a fudisLink href="" [size]="'inherit'" [title]="'non-external link of size inherit'"></a>
+        <a fudisLink href="" [size]="'inherit'" [title]="'non-external link'"></a>
```
RiinaKuu, line 110: All of the `size` properties (with links inside fudis-body-text) are redundant since link inherits BodyText's font-size now.

ngx-fudis/projects/ngx-fudis/src/lib/directives/link/link.directive.stories.ts
```diff
@@ -156,7 +126,7 @@ export const AllCombinations: StoryFn = (args) => ({
 Example.args = {
   title: 'This is an example title',
   external: false,
```
RiinaKuu, line 128: Not an issue of this specific PR, but at some point between `3.2.1` and `4.0.0-rc.0` this story's `external` toggle has broke. It now needs "remount" the view to show the external icon when toggled from the controls.

ngx-fudis/projects/ngx-fudis/src/lib/components/typography/body-text/body-text-theme.scss
```diff
@@ -9,6 +10,10 @@ fudis-body-text {
   &:has(+ fudis-body-text) {
     margin-bottom: spacing.$spacing-xs;
   }
+
+  .fudis-link {
```
RiinaKuu, line 14: Even though this is quite self-explanatory addition for the styles, I would recommend adding a comment here.
 
These `theme` files can get hard to read over time and the changes are always made for some specific workaround or exception --> comments make it easier to keep track what is still relevant.

RiinaKuu (changes requested):

RiinaKuu (approved):

PR #523, opened 2024-12-23
[Label]: Automagically setting Label heights

videoeero:
Jira ticket: https://funidata.atlassian.net/browse/DS-428

This PR introduces new internal `FudisLabelHeightService` tool, which is consumed by Label Component. Service will check each registered Label component and pair them with other Labels which are in the same "row" / vertical position of the view. Then it checks which of the Labels is the heigthest and match other shorter Labels to that, so their related input fields will then align for better UI cohesion.


ngx-fudis/projects/ngx-fudis/src/lib/components/section/section.component.html
```diff
@@ -3,7 +3,7 @@
   fudisGrid
   [id]="id"
   [tabIndex]="-1"
-  [attr.aria-describedby]="_headingId"
+  [attr.aria-labelledby]="_headingId"
```
RiinaKuu, line 6: This fixed the landmark issue in Section Component but quick test with Expandable Component showed that our expandables are not listed in screen reader's (VoiceOver) landmark list. So probably we should change the `aria-describedby` to `aria-labelledby` there since it is using native `section` element as well.

RiinaKuu (changes requested):

RiinaKuu (dismissed):

RiinaKuu (approved):

PR #527, opened 2024-12-30
[Loading Spinner]: New Component

videoeero:
Jira ticket: https://funidata.atlassian.net/browse/DS-248

- Introduces new component: Loading Spinner, which is first Stand Alone component in Fudis
- Tweaked Heading Component, so it is easier to manually focus to native H-element.
- Created Storybook component for Static Components with (mostly) non-interactive components and removed individual Playwright tests for those now in Static Components Playground


RiinaKuu (approved):
Very nice spinner component 🥇 
PR is approved but let's merge this **after** v4.0.0 has been taken from the current main and include this in our next release.

PR #529, opened 2025-01-10
[Label]: Height calculation not working properly

RiinaKuu:
https://funidata.atlassian.net/browse/DS-434

Internal `LabelHeightService` executes automatic label height calculations. It needed to be calculated with `--fudis-rem-multiplier` in order it to scale properly in different Applications.

Additional small fix for `text-emptystate` mixin styles: `font-weight` changed from medium `500` to regular `400`.
Wanted styles can be seen in Typography [Figma file](https://www.figma.com/design/a6G6ZD6W3tlq0qmUkdvdp1/Fudis-DS-Foundations?node-id=26-193&t=DJGEV4hOv9JR4USB-0).


ngx-fudis/projects/ngx-fudis/src/lib/services/dom/label-height.service.ts
```diff
@@ -122,7 +122,8 @@ export class FudisLabelHeightService {
       const height = labelGroups[groupKey].highest;
 
       labelGroups[groupKey].data.forEach((option) => {
-        option.element.style.height = `${height / fontSize}rem`;
+        const remHeight = `${height / fontSize}rem`;
```
marisanity, line 125: Add unit tests to this service, otherwise looks good! 👍

marisanity (approved):

PR #530, opened 2025-01-13
[Icon]: Paperclip icon

RiinaKuu:
https://funidata.atlassian.net/browse/DS-435

- Added new paperclip icon
   - Typically used in cases where attachments are handled
- Updated static components' VR images


marisanity (approved):
LGTM! 👍

PR #536, opened 2025-01-27
[Select, Multiselect]: Usability fixes

RiinaKuu:
https://funidata.atlassian.net/browse/DS-421

SelectOption- and MultiSelectOption element id's are now generated as the html spec requires. Form components including a bug where change event was triggered right after the initialisation is fixed as well.  Also possible memory leak issues handled regarding the observable subscribtions. 

- Changed internal `getNewSelectOptionId` to accept an `label` argument of type `string` and it now returns an id that consists of the parent select id and hashed label. It uses a function `createSelectOptionId` that generates the final option id's by generating a hashed string of the label using `hashLabel`-function.
- Changed internal `_sortSelectedOptionsDOMOrder`-function to implement the use of the new option id's.
- MultiSelect-, Select-, CheckBoxGroup-, Datepicker-, LocalizedTextGroup-, RadioButtonGroup-, TextArea- and TextInput components no longer override the `FormControl`s native `updateValueAndValidity`-function. 
   - Overriding caused an issue with unwanted valueChange events in application side
- Subscriptions are terminated when a `FormControl` inputs change or when the component is destroyed.


ngx-fudis/projects/ngx-fudis/src/lib/components/form/text-input/text-input.component.spec.ts
```diff
@@ -63,9 +63,22 @@ describe('TextInputComponent', () => {
     expect(typeAttribute).toEqual(type);
   }
 
-  it('should create', () => {
+  it('should init the component successfully', () => {
```
marisanity, line 66: I had to cut corners a bit, as the components were quite hard the test because almost every field and function is protected or private. These fields and functions were necessary to test, so I did a workaround so I can assert the variables and functions by disabling the any-rules and getting the needed properties.

MayaMarjut (approved):

PR #537, opened 2025-01-30
[DescriptionList & BodyText]: Lang attribute

MayaMarjut:
https://funidata.atlassian.net/browse/DS-335

Previously screen readers couldn't determine the language used in Description List Item Detail element. Now lang-input, two letter abbreviation, is used in DD element lang attribute. FudisLanguageAbbr type accepts 'fi', 'en'  and 'sv', which are [Official Language Code references](https://www.w3schools.com/tags/ref_language_codes.asp). After changes, screen readers will correctly pronounce these languages.

Since Language Badge Group can be used as part of Description List component or as separately, it was necessary to add lang attribute support also for Body Text component. Since the lang attribute is optional in the Body Text component, the documentation was updated, so that the developers would identify the accessibility challenges and use attribute correctly. If no lang attribute is given, element will naturally inherit current site language.

Added unit test to Description List Item Detail and Body Text, for checking lang attribute.
While updating Description List child component's tests, realised that everything wrapped in `mockFixture.whenRenderingDone().then(()` wasn't really working. Since both components (Description List Item Detail and Description List Item Term) have relative large MockComponents, it was very challenging to test language change and component rendering in this test setup. Eventually decided to remove test 'should remove Language badge group if dd elements with lang do not exists' in Description List Item Term, totally, since it was purely impossible to test. Related test concerning  showing and hiding language options (switching ngIf for `langVisible`), had to be initialised separately. Doing this switch in a one test was impossible.


RiinaKuu (changes requested):

RiinaKuu:
Our PR label categorization only supports single label. 
Improvement overwrites bug fix, so need to decide which one is it.

RiinaKuu (approved):

PR #538, opened 2025-01-31
[AlertGroup]: Screen overflow fix

MayaMarjut:
https://funidata.atlassian.net/browse/DS-402

The screen overflow bug was caused by the overflowing Grid elements. Adjusting grid column suitable for a 320px width screen fixed the issue. Since there is a high change that the client application has a similar kind of problems related to overflowing content on a very small screen - added media query for alert group detecting screen orientation. This will shift between mobile 100 view width and 100% desktop width. Updated VR images and reduced the amount of screenshots.


test/visual-regression/alert-group.spec.ts
```diff
@@ -11,21 +11,27 @@ test("alert group default", async ({ page }) => {
...
+  await page.getByTestId("fudis-alert-4-button").click();
+  await expect(page.getByText("Nothing special here.")).not.toBeVisible();
+  await expect(page.getByText("Yippee Ki-Yay! You were successful!")).not.toBeVisible();
 
   // Open dialog
-  await page.getByTestId("fudis-button-7").click();
-  await page.getByTestId("fudis-alert-3-button").click(); // Dismiss one Alert
+  await page.getByTestId("fudis-alert-6-button").click(); // Dismiss one Alert
+  await page.getByTestId("fudis-button-3").click(); // Add success Alert
+  await page.getByTestId("fudis-button-7").click(); // Open dialog
```
RiinaKuu, line 22: I would suggest adding expects after these button clicks; check that new alert is added and dialog is opened.

I would also add an expect after dialog is closed with Escape keyboard press.

RiinaKuu (changes requested):

RiinaKuu (approved):

PR #546, opened 2025-02-18
[Grid]: Correct the type of columns field

marisanity:
Fixed a bug 

- Removed an redundant variable declaration in `getGridClasses`
- Created interface `FudisDefaultGridProperties` and made that and existing `FudisGridProperties` to extend `FudisGridBase` for getting common properties
- `FudisGridService` is now using and returning `FudisDefaultGridProperties` type of object
- `FudisGridProperties` `columns`-field is now typed correctly and accepting `string` and `number` as well as declared in `GridApiDirective`


ngx-fudis/projects/ngx-fudis/src/lib/directives/grid/gridUtils.ts
```diff
@@ -58,13 +56,7 @@ export const getGridCssValue = (value: number | string, isGridItem?: boolean): s
...
   }
 
-  if (
-    typeof value === 'string' &&
-    value !== 'start' &&
-    value !== 'stretch' &&
-    value !== 'end' &&
-    value !== 'center'
-  ) {
+  if (value !== 'start' && value !== 'stretch' && value !== 'end' && value !== 'center') {
```
marisanity, line 59: There is a check before this evaluation that if value is typeof number, return x. So there is no situation here that value could be anything else than a string.

ngx-fudis/projects/ngx-fudis/src/lib/directives/grid/grid/grid.directive.ts
```diff
@@ -32,7 +32,7 @@ export class GridDirective extends GridApiDirective implements OnInit, OnChanges
   ) {
     super();
 
-    this._element = _gridElement.nativeElement;
+    this._element = this._gridElement.nativeElement;
```
marisanity, line 35: this-keyword added just for clarity

ngx-fudis/projects/ngx-fudis/src/lib/directives/grid/grid/grid.directive.ts
```diff
@@ -60,7 +59,7 @@ export class GridDirective extends GridApiDirective implements OnInit, OnChanges
       /**
        * Get GridServices values
        */
-      this._gridInputProperties.serviceValues = _gridService.getDefaultValues()();
+      this._gridInputProperties.serviceValues = this._gridService.getDefaultValues()();
```
marisanity, line 62: this-keyword added just for clarity

RiinaKuu (commented):
In _Storybook - Services - Grid - setDefaultValues()_ there is a reference for `FudisGridProperties` interface which includes the modified `columns`, and should be updated accordingly.

Otherwise everything looks good 👍 

edit: Also add label for this PR.

RiinaKuu (approved):

PR #551, opened 2025-02-25
[Dialog]: Prevent closing a dialog if the button is disabled

marisanity:
https://funidata.atlassian.net/browse/DS-422

Prevent closing a dialog through `FormSubmitDirective` and `DialogCloseDirective` if the button has been disabled.

### DIFF
--- a/ngx-fudis/projects/ngx-fudis/src/lib/components/dialog/dialog-directives.ts
+++ b/ngx-fudis/projects/ngx-fudis/src/lib/components/dialog/dialog-directives.ts
@@ -94,7 +94,7 @@ export class DialogActionsDirective extends MatDialogActions {
 @Directive({
   selector: '[fudisDialogClose]',
 })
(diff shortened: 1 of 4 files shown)


ngx-fudis/projects/ngx-fudis/src/lib/components/dialog/dialog-directives.ts
```diff
@@ -94,4 +95,19 @@ export class DialogActionsDirective extends MatDialogActions {
 @Directive({
   selector: '[fudisDialogClose]',
 })
-export class DialogCloseDirective extends MatDialogClose {}
+export class DialogCloseDirective extends MatDialogClose {
+  private hostButtonElement: HTMLElement;
```
RiinaKuu, line 99: Since this is a private member, let's add underscore for consistency `_hostButtonElement`.

RiinaKuu (changes requested):

RiinaKuu (approved):
Works as intended, also nice catch to fix the form submit button as well 👍

PR #553, opened 2025-03-01
[Checkbox, CheckboxGroup]: Checkbox value not read correctly by screenreaders on Mozilla Firefox

marisanity:
https://funidata.atlassian.net/browse/DS-454

Styles fixed affecting the readability of checkbox input. Also a fix added regarding the form errors, that were read for the users before the form was actually invalid.


ngx-fudis/projects/ngx-fudis/src/lib/components/form/checkbox-group/checkbox/checkbox.component.html
```diff
@@ -16,11 +16,11 @@
       [name]="_checkboxGroup.id"
       [attr.tabindex]="control.disabled || _checkboxGroup.disabled ? -1 : null"
       [attr.aria-disabled]="control.disabled || _checkboxGroup.disabled || null"
-      [attr.checked]="control.value || null"
```
marisanity, line 19: This is not needed as the input is type checkbox, so it comes with naturally.

PR #554, opened 2025-03-01
[Checkbox, CheckboxGroup]: Checkbox value not read correctly by screenreaders on Mozilla Firefox and Edge

marisanity:
https://funidata.atlassian.net/browse/DS-456

Styles fixed affecting the readability of checkbox input. Also a fix added regarding the form errors, that were read for the users before the form was actually invalid.


ngx-fudis/projects/ngx-fudis/src/lib/components/form/checkbox-group/checkbox/checkbox.component.html
```diff
@@ -16,11 +16,11 @@
       [name]="_checkboxGroup.id"
       [attr.tabindex]="control.disabled || _checkboxGroup.disabled ? -1 : null"
       [attr.aria-disabled]="control.disabled || _checkboxGroup.disabled || null"
-      [attr.checked]="control.value || null"
```
marisanity, line 19: This is not needed as the input is type checkbox, so this is included naturally

ngx-fudis/projects/ngx-fudis/src/lib/components/form/checkbox-group/checkbox/checkbox.component.scss
```diff
@@ -72,13 +72,15 @@
   &__input {
     @include utilities.box-reset;
 
-    appearance: none;
```
marisanity, line 75: You really can't (as long as I'm aware, prove me wrong?) hide the checkbox input with `appearance: none` or any other way where the element completely shrinked to 0x0 pixels. So I used a bit more accessible styling by using `opacity: 0`, it's a nice trick for hiding an element so it's still giving the screen readers the needed input. As there is no need for it to be default width and height as it's hidden anyways, the size is reduced to 1x1 px.

marisanity:
Had to also update the snapshots for the vr tests that contain the checkbox, as there is a 1px difference.

MayaMarjut (approved):

PR #558, opened 2025-03-06
[Popover]: Create a new popover directive

marisanity:
https://funidata.atlassian.net/browse/DS-441

Create new Popover directive, that is a accessible replacement for current existing tooltip.

- Created a popover component that is used by the Popover directive
- Created a popover directive
- Replaced the tooltip to popover in
  - Description List Item Term
  - Datepicker
  - Section
  - MultiSelect
  - Select
  - Button
  - FieldSet
  - Localized Text Group
  - Radio Button Group
  - Checkbox Group
  - Text Input
  - Text Area 
- Removed the tooltip from Language Badge Group


ngx-fudis/projects/ngx-fudis/src/lib/components/popover/popover.component.scss
```diff
@@ -0,0 +1,18 @@
+@use '../../foundations/spacing/tokens.scss' as spacing;
+@use '../../foundations/colors/tokens.scss' as colors;
+@use '../../foundations/typography/mixins.scss' as typography;
+@use '../../foundations/borders/mixins.scss' as borders;
+
+/* stylelint-disable property-disallowed-list */
+/* stylelint-disable unit-disallowed-list */
+
```
marisanity, line 8: Can I avoid these somehow? Why backgroud-color is on a disallowed list?

test/visual-regression/tooltip.spec.ts
```diff
@@ -53,16 +53,3 @@ test("tooltip hover", async ({ page }) => {
   await page.mouse.move(0, 0); // Move mouse away from the tooltip
   await expect(page.locator("mat-tooltip-component")).not.toBeVisible();
 });
-
```
marisanity, line 56: As we have now no tooltips on lang badge group and did not find a suitable documentation page that had enough content to execute this scroll test, I just deleted this test that I wrote a month ago. Tooltip is deprecated anyhow after this feature is merged and removed in the future, so this isn't a huge loss.

ngx-fudis/projects/ngx-fudis/src/lib/components/popover/popover.component.scss
marisanity: And rem as well?

MayaMarjut: Rem values are disallowed for the same reason as we are disallowing hex color values. Basically all style values should come from foundation values. This way we ensure that they are following 8px base. When all sizes have their own spacing token the system is modular and easy to maintain.

MayaMarjut: So in this case when you replace fudis-popover 'background-color' and 'color' with mixins bg-color and text-color from color mixins, you will can get rid off 'stylelint-disable property-disallowed-list',  but yes rem vaues still require to be disabled with 'unit-disallowed-list'

ngx-fudis/projects/ngx-fudis/src/lib/components/form/fieldset/fieldset.component.html
```diff
@@ -28,18 +28,16 @@
...
-
         <fudis-button
-          *ngIf="tooltip"
+          *ngIf="popoverText && popoverTriggerLabel"
           class="fudis-fieldset__legend__main__tooltip-button"
           [size]="'icon-only'"
           [variant]="'tertiary'"
           [icon]="'info-circle'"
           [labelHidden]="true"
-          [label]="tooltip"
```
MayaMarjut, line 39: Since tooltip is removed, should we also remove the extend TooltipApiDirective from fieldset.component.ts? There should be no need to list tooltip props in Storybook excludes, since component is not using them.

ngx-fudis/projects/ngx-fudis/src/lib/utilities/storybook.ts
```diff
@@ -31,6 +31,15 @@ export const excludeEverythingExceptRegex = (array?: string[]): RegExp => {
 
 export const excludeAllRegex: RegExp = /.*/;
 
+/**
+ * Fieldset
+ */
+export const fieldsetExclude: RegExp = excludeRegex([
+  'tooltip',
+  'tooltipPosition',
+  'tooltipToggle',
```
MayaMarjut, line 40: I wonder if these are needed if we just remove the TooltipApi extend from fieldset typescript file

ngx-fudis/projects/ngx-fudis/src/lib/components/form/fieldset/fieldset.component.html
marisanity: The component is using the TooltipApiDirective because popover parameters are there as well so it can't be straight up removed. That's why the tooltip parameters are excluded

ngx-fudis/projects/ngx-fudis/src/lib/utilities/storybook.ts
marisanity: Reply here: https://github.com/funidata/fudis/pull/558#discussion_r2002742239

ngx-fudis/projects/ngx-fudis/src/lib/directives/popover/popover.mdx
```diff
@@ -0,0 +1,55 @@
...
+
+Popover shows contextual help or information about specific element when user clicks on it or presses enter/space key.
+
+## Usage Guideline
+
+To add popover to any HTML element, add `fudisPopover` directive to it with a string text property `popoverText`. You can also specify a desired position for the popover with optional property `popoverPosition`. The default value is `below`. If the popover does not have enough space in the viewport, it will fallback to a position where is enough space.
+
+Popover is triggered on mouse click and on keyboard enter or space press. Popover content is announced for screen reader users when the element where the popover is anchored is triggered. Popover content is hidden only when the anchored element is triggered again, page content is scrolled or when focus is moved outside the popover with mouse click or keypress.
+
+The interactive element where the popover is anchored to, should always have a aria-label, that describes the what is the purpose of the element.
```
MayaMarjut, line 18: I wonder if this paragraph is littlepit misleading. Interactive element needs to have a role, like native button has a role="button". This way assistive technology will know how to trigger popover. If popover is added to a element that contains text, that text will work as a suitable label. There is no need to add a separate aria-label for that case, since aria-label will override visible text content / or aria-labelledby attribute. When popover is used with information icon, adding aria-label would be mandatory.

ngx-fudis/projects/ngx-fudis/src/lib/directives/popover/popover.directive.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+  animations: [],
+  template: `<ng-container #container>
+    <button fudisPopover [popoverText]="popoverText" [popoverPosition]="popoverPosition"
+      >Open a popover</button
+    >
+  </ng-container>`,
+})
+class MockPopoverComponent {
+  @ViewChild('container', { read: ViewContainerRef }) container!: ViewContainerRef;
+  @Input() popoverPosition: FudisPopoverPosition = 'below';
```
MayaMarjut, line 21: You could add a test that loops through all popoverPositions

test/visual-regression/popover.spec.ts
```diff
@@ -0,0 +1,48 @@
+import test, { expect, Page } from "@playwright/test";
+
```
MayaMarjut, line 2: Could have a test for clicking the document and checking that popover is not visible

test/visual-regression/tooltip.spec.ts
```diff
@@ -22,26 +22,26 @@ test("tooltip toggle", async ({ page }) => {
   await page.mouse.click(0, 0); // Click away from the tooltip
```
MayaMarjut, line 22: This test could be added to popover as well

ngx-fudis/projects/ngx-fudis/src/lib/directives/popover/popover.mdx
marisanity: I think I got caught up in the fact that most of our components have just icon buttons in them, you are right. I removed that line. I have anyways documented that in the fudis components that require the `popoverTriggerLabel`.

test/visual-regression/popover.spec.ts
marisanity: Did a test using the click mentioned here https://github.com/funidata/fudis/pull/558#discussion_r2003073625

ngx-fudis/projects/ngx-fudis/src/lib/directives/popover/popover.directive.spec.ts
marisanity: I did an visual assertion instead that asserts below and right 👍

test/visual-regression/tooltip.spec.ts
marisanity: https://github.com/funidata/fudis/pull/558#discussion_r2003328937

ngx-fudis/projects/ngx-fudis/src/lib/directives/popover/popover.directive.spec.ts
MayaMarjut: All directions should be tested. Also, it would de nice that we wouldn't need so many pictures of popovers. So since it's not possible to have a single picture of all popovers visible, I might suggest that these positions inputs would be tested in unit tests.

PR #559, opened 2025-03-10
[Dependencies]: Update Angular 19

RiinaKuu:
https://funidata.atlassian.net/browse/DS-445

Breaking:
- Updated peerDependencies:
    `"@angular/cdk"`: `"^19.2.1"`,
    `"@angular/common"`: `"^19.2.0"`,
    `"@angular/core"`: `"^19.2.0"`,
    `"@angular/material"`: `"^19.2.3"`
    
Info:
- From now on all new components and directives are standalone by default, otherwise have to declare `standalone: false`.
- Removed `ng-mocks` from devDependencies. We shouldn't have any reason to mock our components in our unit tests.
- Fixed miscellaneous failing tests and poor coding where desired variable didn't update


MayaMarjut:
dialog-directive.spec.ts and backend-simulation-base.directive.ts have still standalone: true. Are these intentional?

RiinaKuu:
> dialog-directive.spec.ts and backend-simulation-base.directive.ts have still standalone: true. Are these intentional?

Both of these have been made as standalone from the beginning intentionally, so it is okay. There are plenty of other example components as well, that are made for Storybook and are intentionally started to use standalone since that is the wanted way now.

RiinaKuu:
> dialog-directive.spec.ts and backend-simulation-base.directive.ts have still standalone: true. Are these intentional?

Not sure why the automatic migration didn't catch these ones. Will delete since they should be irrelevant.

MayaMarjut (approved):

PR #560, opened 2025-03-10
[DEPENDENCIES]: Storybook upgrade v8.6.3

MayaMarjut:
https://funidata.atlassian.net/browse/DS-444

Updated Storybook to a new minor. Since Storybook >= 8.0 does not support GFM markup for tables in MDX documentation. [Followed Storybook documentation](https://storybook.js.org/docs/writing-docs/mdx#markdown-tables-arent-rendering-correctly) on installing GFM plugin.

Now Storybook has a new 'Interaction' Tab without content. Decided to keep '@storybook/addon-interactions' since it was already included in package.json, but could be probably removed.


ngx-fudis/package.json
```diff
@@ -30,6 +30,7 @@
     "@angular/platform-browser-dynamic": "^18.2.13",
     "@angular/router": "^18.2.13",
     "eslint-config-prettier": "^10.0.1",
+    "remark-gfm": "^4.0.1",
```
RiinaKuu, line 33: Isn't this plugin specifically meant for **markdown table support** and since we are not using any, why do we need it?

MayaMarjut: We are using tables in our mdx files. This plugging ads support for markdown and <table> -tags

MayaMarjut: Seemed to work without, maybe some update issue? Removed Markdown tag from InformationSecurityManagement doc page

RiinaKuu (approved):

PR #561, opened 2025-03-17
[Breadcrumbs, Grid]: Add and clean up tests

RiinaKuu:
- Removed unnecessary effect in Breadcrumbs Component, there were no signals used
- Existing VR tests for Grid and Grid Item were sufficient, these components don't need unit test since their main functionality is not feasible to test in Jest.
- Added missing unit tests for Grid Utils

### DIFF
--- a/ngx-fudis/projects/ngx-fudis/src/lib/components/breadcrumbs/breadcrumbs.component.ts
+++ b/ngx-fudis/projects/ngx-fudis/src/lib/components/breadcrumbs/breadcrumbs.component.ts
@@ -3,7 +3,6 @@ import {
   Input,
   ViewEncapsulation,
   ChangeDetectionStrategy,
-  effect,
 } from '@angular/core';
 import { FudisTranslationService } from '../../services/translation/translation.service';
 import { FudisIdService } from '../../services/id/id.service';
@@ -22,10 +21,7 @@ export class BreadcrumbsComponent {
     private _idService: FudisIdService,
   ) {
     this._id = this._idService.getNewParentId('breadcrumbs');
-
-    effect(() => {
-      this._breadcrumbsPrefix.next(this._translationService.getTranslations()().BREADCRUMBS.PREFIX);
-    });
+    this._breadcrumbsPrefix.next(this._translationService.getTranslations()().BREADCRUMBS.PREFIX);
   }
 
   /**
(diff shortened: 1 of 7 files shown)


ngx-fudis/projects/ngx-fudis/src/lib/components/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -23,10 +17,7 @@ export class BreadcrumbsComponent {
     private _idService: FudisIdService,
   ) {
     this._id = this._idService.getNewParentId('breadcrumbs');
-
-    effect(() => {
-      this._breadcrumbsPrefix.next(this._translationService.getTranslations()().BREADCRUMBS.PREFIX);
-    });
```
marisanity, line 29: Laukeskohan tää? Ku tossa on siis toi `this._translationService.getTranslations()().BREADCRUMBS.PREFIX`, joka on siis signaali?

marisanity: I'm not sure if this is the case, but I'd check that! 👍

RiinaKuu: This is a static language key from our internal translations and after component initialization it will not change, works fine even without effect.

ngx-fudis/projects/ngx-fudis/src/lib/directives/grid/grid-utils.spec.ts
```diff
@@ -28,6 +31,13 @@ describe('GridUtils', () => {
 
       expect(correctReturnString).toBe('1/-1');
 
+      fudisGridAlignItemsArray.forEach((value) => {
```
MayaMarjut, line 34: Since all of these test functions are lengthy, I would organise them to follow the same order as presented in ts file. Etc, put 'getGrid Classes' as the first test, improves readability.

MayaMarjut:
Can you recheck the link for 'Auto, Auto-Fill and Auto-Fit' in Grid documentation page. I think the url id is broken.

MayaMarjut (approved):

PR #563, opened 2025-03-22
[Tooltip Directive]: DEPRECATION flag

MayaMarjut:
https://funidata.atlassian.net/browse/DS-451

Added deprecation flag to Tooltip Directive. Documented the decision of directive removal on major version 7.0. Also removed autoDocs: true from storybook main.js, which was included in Storybook upgrade. Felt that is was redundant.

Note: Wasn't sure if this is a chore.

### DIFF
--- a/ngx-fudis/.storybook/main.js
+++ b/ngx-fudis/.storybook/main.js
@@ -27,7 +27,6 @@ export const staticDirs = [
 ];
 export const docs = {
   defaultName: "Documentation",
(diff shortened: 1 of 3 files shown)


RiinaKuu (commented):
Just a thought, but should the deprecation warning be visible in the Tooltip's menu heading also?
Screenshot 2025-03-24 at 12 15 24

RiinaKuu (approved):

PR #566, opened 2025-03-28
[Expandable, Fieldset, Footer, Form, Section]:Content projection directives unit tests

MayaMarjut:
https://funidata.atlassian.net/browse/DS-264

Created a new type fudisFieldsetActionsAlign.

Improved content directives unit test to check host classes.

### DIFF
--- a/ngx-fudis/projects/ngx-fudis/src/lib/components/expandable/expandable-content.directive.spec.ts
+++ b/ngx-fudis/projects/ngx-fudis/src/lib/components/expandable/expandable-content.directive.spec.ts
@@ -16,6 +16,6 @@ describe('ExpandableActionsDirective', () => {
   it('should have host class', () => {
     const directive = new ExpandableActionsDirective();
 
-    expect(directive['_classes']).toBe('fudis-expandable-actions');
+    expect(directive['_hostClass']).toBe('fudis-expandable-actions');
   });
 });
--- a/ngx-fudis/projects/ngx-fudis/src/lib/components/expandable/expandable-content.directive.spec.ts
+++ b/ngx-fudis/projects/ngx-fudis/src/lib/components/expandable/expandable-content.directive.spec.ts
@@ -1,0 +1,22 @@
+import { TestBed } from '@angular/core/testing';
+import { ExpandableActionsDirective } from './expandable-content.directive';
+
+describe('ExpandableActionsDirective', () => {
+  beforeEach(() => {
+    TestBed.configureTestingModule({
+      declarations: [  
+        ExpandableActionsDirective ]
+    });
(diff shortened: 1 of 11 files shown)


RiinaKuu:
Overall I think these tests are enough for such content projection directives, but I feel like there are unnecessary and/or self-evident **comments that could be removed** in all of the four directive files (`expandable-content.directive.ts`, `footer-content.directive.ts`, `form-content.directive.ts` and `fieldset-content.directive.ts`). The only comment I would **leave as is**, is Fieldset’s align comment.

For the sake of cohesiveness:
- rename `_classes` to `_hostClass` in `expandable-content.directive.ts`
- change three `public hostClass` to `private _hostClass` in `expandable-content.directive.ts`

Also the naming is a bit confusing, e.g. `expandable-content.directive.ts` which then includes **ExpandableActionsDirective** and **ExpandableContentDirective**. This is just a personal opinion but I would prefer if it differs from the ContentDirective itself. Could it just be `expandable-contents.directive.ts` or is there some better synonyms for content 🤔

RiinaKuu (changes requested):
We might have missed one content directive when creating the Jira-ticket, `section-content.directive.ts` is still missing basic tests. Also it has unnecessary comments (or if you want to leave the comments, all the `_hostClass` should have the same comment in all the content directives) and `public hostClass` should be fine to change to `private _hostClass`.

MayaMarjut:
> Overall I think these tests are enough for such content projection directives, but I feel like there are unnecessary and/or self-evident **comments that could be removed** in all of the four directive files (`expandable-content.directive.ts`, `footer-content.directive.ts`, `form-content.directive.ts` and `fieldset-content.directive.ts`). The only comment I would **leave as is**, is Fieldset’s align comment.
> 
> For the sake of cohesiveness:
> 
> * rename `_classes` to `_hostClass` in `expandable-content.directive.ts`
> * change three `public hostClass` to `private _hostClass` in `expandable-content.directive.ts`
> 
> Also the file naming is a bit confusing, e.g. `expandable-content.directive.ts` which then includes **ExpandableActionsDirective** and **ExpandableContentDirective**. This is just a personal opinion but I would prefer if the file name differs from the ContentDirective itself. Could it just be `expandable-contents.directive.ts` or is there some better synonyms for content 🤔

- I understand your point, but then I should need to change all the directives names `expandable-contents.directive.ts`, `footer-contents.directive.ts`, `form-contents.directive.ts` and `fieldset-contents.directive.ts`) to pluriel. This is again a personal opinion, but 'contents' doesn't fit in my mouth and seems like a typo.

RiinaKuu (approved):

PR #573, opened 2025-04-04
[TextInput]: New HTML autocomplete attribute

RiinaKuu:
https://funidata.atlassian.net/browse/DS-466

- Added two new inputs to TextInput component
   - `autocompleteFill` -  adds HTML autocomplete
   - `name` - adds HTML name attribute
- Updated unit tests and documentation


ngx-fudis/projects/ngx-fudis/src/lib/components/form/text-input/readme.mdx
```diff
@@ -17,6 +17,17 @@ Text Input type options are `text` (default), `email`, `number`, `password`, `te
 
 Necessary properties for Text Input are `label` and `control`.
 
+### Autocomplete
+
+`autocompleteFill` will add HTML autocomplete attribute to the input which will then provide automated assistance in filling out form field values.
```
MayaMarjut, line 22: Since autocomplete is native HTML attribute, could we just use the same name? At least we are following this naming convention with rest of native attributes.

test/visual-regression/form-dynamic-validators.spec.ts
```diff
@@ -46,7 +46,7 @@ test("form dynamic inputs", async ({ page }) => {
   await page.getByTestId("fudis-button-9").click(); // remove required number validator
   await page.getByTestId("fudis-button-12").click(); // remove required date validator
   await page.getByTestId("fudis-button-15").click(); // remove required option validator
-  await page.getByText("Summer holidays").click(); // Check the first checkbox to remove error messages
+  await page.getByTestId("fudis-checkbox-group-1-item-1").check(); // Check the first checkbox to remove error messages
```
MayaMarjut, line 49: 'click()'

RiinaKuu: In Playwright documentation `.check()` is used for checkboxes and radio-buttons: https://playwright.dev/docs/input#checkboxes-and-radio-buttons

But it also suggests using `.setChecked(true)`. If you test that, does it still fail locally for you?

ngx-fudis/projects/ngx-fudis/src/lib/components/form/text-input/readme.mdx
RiinaKuu: I was trying to add some differentiation to our Select with autocomplete but maybe it's in vain. Will rename it to match the native attribute.

test/visual-regression/form-dynamic-validators.spec.ts
RiinaKuu: Actually it failed locally for me too now with `.check()` and passed with `.setChecked(true)` so I changed it.

MayaMarjut (approved):

PR #574, opened 2025-04-07
[TabNavigation]: Create a tab navigation component

marisanity:
Create a tab navigation component that supports routing.

DRAFT, still missing:
- Documentation
- Tests


MayaMarjut:
I honestly think that this is not accessible solution. Everything with this opacity overflow from 0 to 100 is considered problematic. I think the whole design could be improved, specially the choice of icons (chevron-ring-fill) in this case. The whole background of the icon should be white, so it wouldn't seem that the content is flowing underneath the icon, but instead of some white box.
Screenshot 2025-04-14 at 9 12 18

MayaMarjut:
All tab elements should have aria-controls value that refers to the associated tabpanel element. It would be nice to have an storybook example where tab panels are actually changing. Is there a way we could achieve this?

ngx-fudis/projects/ngx-fudis/src/lib/components/tab-navigation/tab-navigation-bar.component.html
```diff
@@ -34,19 +35,20 @@
...
-      tabindex="-1"
-      [class]="
+      [className]="
         'fudis-tab-navigation-bar__button-container__button fudis-tab-navigation-bar__button-container__button--right'
       "
+      tabindex="-1"
+      variant="tertiary"
+      [labelHidden]="true"
+      icon="chevron-ring-fill"
+      label="Scroll view right"
```
MayaMarjut, line 47: These labels may need a default translation, even though they are hidden from screen readers.

ngx-fudis/projects/ngx-fudis/src/lib/components/tab-navigation/tab-navigation-bar.component.html
```diff
@@ -3,21 +3,22 @@
...
-      tabindex="-1"
-      [class]="
+      [className]="
         'fudis-tab-navigation-bar__button-container__button fudis-tab-navigation-bar__button-container__button--left'
       "
+      tabindex="-1"
+      variant="tertiary"
+      [labelHidden]="true"
+      icon="chevron-ring-fill"
+      label="Scroll view left"
```
MayaMarjut, line 15: These labels may need a default translation, even though they are hidden from screen readers.

MayaMarjut:
> I honestly think that this is not accessible solution. Everything with this opacity overflow from 0 to 100 is considered problematic. I think the whole design could be improved, specially the choice of icons (chevron-ring-fill) in this case. The whole background of the icon should be white, so it wouldn't seem that the content is flowing underneath the icon, but instead of some white box. Screenshot 2025-04-14 at 9 12 18

I edited the design to be more accessible, with the mutual agreement with designers. Also quote: "Using a ::before pseudo-element to add content is discouraged, as it is not reliably accessible to screen readers" (https://developer.mozilla.org/en-US/docs/Web/CSS/::before).

MayaMarjut:
Would be good to add VR test for both scroll buttons showing on mobile view

marisanity:
> All tab elements should have aria-controls value that refers to the associated tabpanel element. It would be nice to have an storybook example where tab panels are actually changing. Is there a way we could achieve this?

I'll implement the aria-controls 👍 

But I would not put any kind of router to storybook. It would be nice, but I don't think it's a good idea. If we make a tab that shows fixed content (that would be a seperate component) and does not use a router, then it is possible to preview the tabs. But I think the code examples are enough for the implementation. I can also add an example of the routing function to clarify the usage if needed.

ngx-fudis/projects/ngx-fudis/src/lib/components/tab-navigation/tab-navigation.mdx
```diff
@@ -0,0 +1,108 @@
...
+  [id]="'fudis-tab-navigation-1'"
+  [variant]="'primary'"
+  [panel]="navigationTabPanel">
+  ...
+</fudis-tab-navigation-bar>
+```
+
+### Tab Navigation Tab
+
+Nested inside the `TabNavigationBarComponent` you have to add content, that are the tabs. You can choose to use `button`- or `a`-tags, mixing them is also acceptable. These require a `fudis-tab-navigation-tab` attribute and two props, `id` and `active`. **You have to provide the event handlers or/and attributes you need to implement the routing logic yourself.**
```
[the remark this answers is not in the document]
marisanity: Ah I see! But I kind of want to support the possibility if router is needed to use with an a-tag... Is it alright, that I just add support for space for the tab that is constructed with the anchor element? So Enter and space work always no matter is it a button- or an a-tag?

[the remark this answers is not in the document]
marisanity: I implemented this, but check it out and see if it is acceptable

ngx-fudis/projects/ngx-fudis/src/lib/components/tab-navigation/tab-navigation.mdx
```diff
@@ -88,9 +90,40 @@ This example is written using angular router as an reference. **However, you can
...
+</div>
+```
+
+It is also important, that the navigation bar and the panel are located next to each other as siblings, even if they are in separate components.
+
 ## Accessibility
 
 - The active tab is marked with `aria-selected` and the ones that are inactive, have a tabindex of -1, so the user can skip the tab listing with tab. Navigation between the tabs is done by left- and right arrow keyboard buttons.
+- The tabs are referencing the tab panel with `aria-controls`
 - Tab panel content is linked to the corresponding tab with `aria-labelledby`
```
MayaMarjut, line 132: Add a comma on rest of these lines.

PR #575, opened 2025-04-08
[Fudis Core]: Create Core project

RiinaKuu:
https://funidata.atlassian.net/browse/DS-468

- Create separate core project inside Fudis
- Copy all SCSS tokens and mixins from Foundations
   - Note: Foundations Storybook documentation files were not copied since Storybook installation will come in a separate PR


core/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "fudis-core",
+  "version": "1.0.0",
+  "description": "Core styles and foundations for Fudis Design System",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
```
RiinaKuu, line 7: Do we want to run some tests in this project? If so, what?

core/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "stylelint": "stylelint '**/*.{css,scss}'",
+    "stylelint-fix": "stylelint '**/*.{css,scss}' --fix"
```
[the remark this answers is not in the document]
RiinaKuu: Okay, I was missing double dash before the flag itself and will remove this script. Correct usage is then  `npm run stylelint -- --fix` and it is probably more beneficial to use it in this way (instead of pre-made script) so that one can add whatever flags they need.

joonashak:
Henlo 👋 

I would look at hoisting the `.editorconfig` and Stylelint stuff up to repo root. Prettier and ESLint tend to work best when there is only one instance of them in the repository – I don't see why this wouldn't be the case with Stylelint, too.

Another benefit is that the setup becomes much simpler and easier to maintain as one set of configuration applies to the whole code base instead of having multiple configuration for the same tool, which tend to differ as time goes by.

core/package.json
marisanity: This could be a seperate ticket? We could check what is available/recommended for style libraries if needed. And of course visual regression could be used in storybook at least.

core/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "@funidata/fudis-core",
+  "version": "1.0.0",
+  "description": "Core styles and foundations for Fudis Design System",
+  "scripts": {
```
marisanity, line 5: Also the styles need to be exported/exposed, as now you cannot access the design tokens and mixins from other libraries.

marisanity: This includes an entry point that is needed as well.

joonashak: I recommend setting up a test project using `install-local` to verify that the setup actually works. While using the `"package": "file:..."` type of dependency specification is easier, it does not actually install the package and some things, such as peer dependency resolution, are easy to miss.

You can take a look, e.g., at https://github.com/joonashak/nestjs-clone-bay/blob/main/dev/package.json for an example.

joonashak: If you wanna be real robust, I'd go as far as committing the test project to repo and writing a simple smoke-test for the CI to ensure that the package works correctly before publishing.

joonashak: Also, when using `install-local` in a test project you also get to see how the imports actually look like. I'm not sure, but I have a suspicion that with the current file structure, the imports may end up having weird forms such as `import "@funidata/fudis-core/src/foundations/borders/mixins.scss";` which is quite verbose. This is easy to test and optimize once you have a test project up.

core/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "stylelint": "stylelint '**/*.{css,scss}'"
+  },
```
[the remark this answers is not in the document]
joonashak: Maybe [sass](https://www.npmjs.com/package/sass)?

core/package.json
marisanity: We also need an entry file. Also alternative to files (don't know which is better, @joonashak give your opinion on this too) could be exports:
`exports: {
    ".": {
      "sass": "./_index.scss",
    },
}`

As this also gives a possibility for fine grained imports:

`
"exports": {
  ".": "./_index.scss",
  "./tokens": "./tokens/_tokens.scss",
  "./mixins": "./mixins/_mixins.scss",
  "./helpers": "./helpers/_helpers.scss",
}
`

Of course the files could include a directory which would contain everything we want to export and leave the rest outside

joonashak: `exports` is not an alternative for `files`, it's an alternative for `main`. I'm not sure if `main` can be used as an export mechanism here since there is no JavaScript that is exported.

Or is there? I'm not sure if this should be set up as SCSS files directly included in the package payload, as it is now, or rather the SCSS compiled to JS and that released as the package payload... 🤔 I think this should be thoroughly tested with, e.g., `install-local` before moving forward. I have no experience with SASS/SCSS.

joonashak: Also, would probably be a good idea to search for legit libraries that do this kind of thing with SCSS and review how they are packaged.

core/package.json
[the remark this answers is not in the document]
marisanity: Yes I meant sass, sorry about that 😆

core/package.json
marisanity: Main can surely be used with sass files as well. But in summa summarum; I think now it is time for `install-local` and test to figure out what combination is the best and working solution for exposing the files. Also a seperate matter, but create a _index.scss and similar files for mixins and tokens (_mixins.scss, _tokens.scss) and forward all related files from there.

marisanity: One additional comment regarding the package.json configurations:

["main"](https://nodejs.org/api/packages.html#main) - The default module when loading the package, if exports is not specified, and in versions of Node.js prior to the introduction of exports.

This is how main is described in node.js documentation and here is the exports field:

["exports"](https://nodejs.org/api/packages.html#exports) - Package exports and conditional exports. When present, limits which submodules can be loaded from within the package.

So if I understood this right, the main might not be required. But yeah, this just needs some deep diving to the configuration documentation to see what is works out for the best. Some of the documentation is in Node.js and some is found in NPM. But as @joonashak commented, it is a good idea to check also what kind of configurations other libraries use. But if there are not any good references, then it is just about testing to see what works out the best.

joonashak: Apparently `exports` takes precedence over `main` if supported. `main` can be included for backwards compatibility. `exports` was introduced in July 2019 so I'm guessing we don't really need it anymore.

https://stackoverflow.com/a/79383294/7676920

core/package.json
RiinaKuu: Created own ticket for this: https://funidata.atlassian.net/browse/DS-487

core/package.json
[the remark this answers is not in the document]
RiinaKuu: Discussed about this and came to the conclusion that we should not have style compiler as peer dependency as user application can choose whichever preprocessor they want to use.

RiinaKuu:
Tested current set up successfully with recommended `install-local`.
The imports in application can be used e.g. like this: `@use '@funidata/fudis-core/mixins' as mixins;` which will expose all available mixins.

core/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "@funidata/fudis-core",
+  "version": "1.0.0",
```
joonashak, line 3: Not saying that this is incorrect _per se_ but since this PR neither includes a publishing mechanism to NPM nor integrates `fudis-core` with `ngx-fudis`, I'd err to the side of caution and drop the version to `0.0.1`. That way there will be no accidental release of v1 before we know that everything works.

package.json
```diff
@@ -17,7 +17,9 @@
     "test:ui": "find test/visual-regression -name \\*-darwin.png -type f -delete && cd test && npm start",
     "build": "docker build -t fudis:latest ./ngx-fudis && docker cp $(docker create fudis:latest):/usr/src/app/dist ./ngx-fudis/dist",
     "lint": "cd ngx-fudis && npm run lint",
-    "stylelint": "cd ngx-fudis && npm run stylelint",
+    "stylelint:ngx-fudis": "cd ngx-fudis && npm run stylelint",
+    "stylelint:core": "cd core && npm run stylelint",
+    "stylelint": "concurrently \"npm run stylelint:ngx-fudis\" \"npm run stylelint:core\"",
```
joonashak, line 22: Now that Stylelint is hoisted to repo root, there is no need for multiple linting commands (or `concurrently`).

- The linting command in this file should be `"stylelint": "stylelint '**/*.{css,scss}'"`. This lints the whole project in one go.
-  In `core/package.json` and `ngx-fudis/package.json`:
  - Stylelint and related packages should be uninstalled.
  - Linting commands should be removed.
- `concurrently` should be removed from `package.json`.

RiinaKuu: Nice! I was too blind for these stylelint commands and didn't even realize they weren't relevant anymore.

core/package.json
RiinaKuu: Agreed 👍

joonashak (approved):
🚀

PR #576, opened 2025-04-08
[RadioButtonGroup,CheckboxGroup,Fieldset]: Add help text to Fieldset legend

MayaMarjut:
https://funidata.atlassian.net/browse/DS-459

With NVDA, aria-describedby fieldset value wasn't working in Radio Button Group or Checkbox Group component. (worked with VoiceOver). Screen readers weren't notified of the help text in Guidance when user focused group component. 

In this pull request added Fieldset logic to determine when parent component is either radio button group or a checkbox group, and  then place that component's Guidance help text as part of the Fieldset legend.

Wrote supporting unit tests for related components. Edited documentation concerning accessibility. 

Also removed 'left:0' from CSS clip method since it's unnecessarily throws screen reader focus to edge of screen when focusing on hidden elements.

Note! These changes won't fix the fact that Guidance error messages are not read to screen reader, in these related group components, in a appropriate phase.


ngx-fudis/projects/ngx-fudis/src/lib/components/form/fieldset/fieldset.component.ts
```diff
@@ -182,6 +190,10 @@ export class FieldSetComponent
 
       this._resizeObserver.observe(this._fieldsetLegend.nativeElement);
     }
+
+    if (['fudis-radio-button-group', 'fudis-checkbox-group'].some((id) => this.id.includes(id))) {
```
RiinaKuu, line 194: Change this id dependant check to the one where we just check if Fieldset has CheckboxGroup or RadioButtonGroup component as its parent.

- Add CheckboxGroup and RadioButtonGroup components to constructor with `@Host()` and `@Optional()` decorators.
- Change this if to check `this._parentCheckBoxGroup || this._parentRadioButtonGroup`

RiinaKuu (changes requested):

ngx-fudis/projects/ngx-fudis/src/lib/components/form/fieldset/fieldset.component.ts
RiinaKuu: This (of course) caused import cycles so we need to do it through ElementRef.

RiinaKuu (approved):

PR #578, opened 2025-04-11
[Guidance]: Unify maxLength indicator screen reader prompts

MayaMarjut:
https://funidata.atlassian.net/browse/DS-447

- Unify screen reader functionality so that it works similarly in LocalizedTextGroup and basic text inputs. 
    - _"x/xx characters used"_ is now always read in its own block after helpText and not within helpText. 
     - Threshold warning functionality stays the same.

- Added suggestion for Dialog's documentation about heading level

### DIFF
--- a/ngx-fudis/projects/ngx-fudis/src/lib/components/form/guidance/guidance.component.html
+++ b/ngx-fudis/projects/ngx-fudis/src/lib/components/form/guidance/guidance.component.html
@@ -85,7 +85,8 @@
       <ng-container *ngIf="helpText">{{ helpText }}</ng-container>
       <ng-container *ngIf="control && maxLength">
         <span class="fudis-visually-hidden"
-          >{{ control.value?.length || 0 }}/{{ maxLength }} {{ _maxLengthText | async }}</span
+          >&nbsp; {{ control.value?.length || 0 }}/{{ maxLength }}
+          {{ _maxLengthText | async }}</span
         >
       </ng-container>
     </p>
@@ -122,7 +123,8 @@
         <p
           class="fudis-guidance__character-limit-indicator__alert fudis-visually-hidden"
           role="alert"
-          >{{ singleControl.value.value?.length || 0 }}/{{ maxLength }}
+        >
+          &nbsp; {{ singleControl.value.value?.length || 0 }}/{{ maxLength }}
           {{ _maxLengthText | async }}</p
         >
       </ng-container>
@@ -137,6 +139,6 @@
   "
 >
   <p class="fudis-guidance__character-limit-indicator__alert fudis-visually-hidden" role="alert"
-    >{{ control.value?.length || 0 }}/{{ maxLength }} {{ _maxLengthText | async }}</p
+    >&nbsp; {{ control.value?.length || 0 }}/{{ maxLength }} {{ _maxLengthText | async }}</p
   >
 </ng-container>
--- a/ngx-fudis/projects/ngx-fudis/src/lib/components/form/guidance/guidance.component.html
+++ b/ngx-fudis/projects/ngx-fudis/src/lib/components/form/guidance/guidance.component.html
@@ -83,36 +83,30 @@
 
     <p *ngIf="helpText || (control && maxLength)" class="fudis-guidance__help-text">
       <ng-container *ngIf="helpText">{{ helpText }}</ng-container>
-      <ng-container *ngIf="control && maxLength">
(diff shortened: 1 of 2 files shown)


RiinaKuu (commented):
If we want to add space between the two sentences (if I understood correct that it bothers you), can't we use `&nbsp;` before the screen reader text?

MayaMarjut:
> If we want to add space between the two sentences (if I understood correct that it bothers you), can't we use `&nbsp;` before the screen reader text?

I thought I tried this, but maybe in a wrong space.

ngx-fudis/projects/ngx-fudis/src/lib/components/form/guidance/guidance.component.html
```diff
@@ -85,13 +85,15 @@
       <ng-container *ngIf="helpText">{{ helpText }}</ng-container>
       <ng-container *ngIf="control && maxLength">
         <span class="fudis-visually-hidden"
-          >{{ control.value?.length || 0 }}/{{ maxLength }} {{ _maxLengthText | async }}</span
+          >&nbsp; {{ control.value?.length || 0 }}/{{ maxLength }}
```
RiinaKuu, line 88: It seems that we have two different implementations for reading the maxLength (`LocalizedTextGroup` vs. basic `TextArea`).

In TextArea it is read **within** the input helpText where as in LocalizedTexGroup it is read **after** helpText. Since it probably would be best to keep the flow same, the easiest way would be deleting this whole `ng-container`... (continued in next comment)

ngx-fudis/projects/ngx-fudis/src/lib/components/form/guidance/guidance.component.html
```diff
@@ -85,13 +85,15 @@
...
+          {{ _maxLengthText | async }}</span
         >
       </ng-container>
     </p>
   </div>
   <ng-container *ngIf="maxLength && control">
     <small
+      aria-hidden="true"
```
RiinaKuu, line 96: Remove this aria-hidden and add visually hidden text after the maxLength indicator, like so:

```
      {{ control.value?.length || 0 }}/{{ maxLength
      }}<span class="fudis-visually-hidden">{{ _maxLengthText | async }}</span>
```

ngx-fudis/projects/ngx-fudis/src/lib/components/form/guidance/guidance.component.html
```diff
@@ -104,6 +106,7 @@
     <ng-container *ngFor="let singleControl of formGroup?.controls | keyvalue">
       <small
         *ngIf="singleControl.key === selectedOption"
+        aria-hidden="true"
```
RiinaKuu, line 109: Remove this aria-hidden and add visually hidden text after the maxLength indicator, like so:

```
        {{ singleControl.value.value?.length || 0 }}/{{ maxLength
        }}<span class="fudis-visually-hidden">{{ _maxLengthText | async }}</span>
```

ngx-fudis/projects/ngx-fudis/src/lib/components/form/guidance/guidance.component.html
```diff
@@ -120,7 +123,8 @@
         <p
           class="fudis-guidance__character-limit-indicator__alert fudis-visually-hidden"
           role="alert"
-          >{{ singleControl.value.value?.length || 0 }}/{{ maxLength }}
+        >
+          &nbsp; {{ singleControl.value.value?.length || 0 }}/{{ maxLength }}
```
RiinaKuu, line 127: Remove unnecessary `&nbsp;`

ngx-fudis/projects/ngx-fudis/src/lib/components/form/guidance/guidance.component.html
```diff
@@ -135,6 +139,6 @@
   "
 >
   <p class="fudis-guidance__character-limit-indicator__alert fudis-visually-hidden" role="alert"
-    >{{ control.value?.length || 0 }}/{{ maxLength }} {{ _maxLengthText | async }}</p
+    >&nbsp; {{ control.value?.length || 0 }}/{{ maxLength }} {{ _maxLengthText | async }}</p
```
RiinaKuu, line 142: Remove unnecessary `&nbsp;`

RiinaKuu (changes requested):

RiinaKuu (approved):

PR #580, opened 2025-04-23
[Core]:Install Storybook

MayaMarjut:
https://funidata.atlassian.net/browse/DS-469

- Created button example (removed styles related to icon button)
- Installed Storybook plugins separately '@storybook/addon-storysource' to display code tab and "@storybook/addon-a11y" to display accessibility tab. 
- Uploaded fonts
- Created Docker file


core/src/components/button/button.scss
```diff
@@ -0,0 +1,72 @@
+@use '../../foundations/colors/mixins.scss' as colors;
+@use '../../foundations/focus/mixins.scss' as focus;
+@use '../../foundations/borders/mixins.scss' as borders;
+@use '../../foundations/typography/mixins.scss' as typography;
+@use '../../foundations/spacing/tokens.scss' as spacing;
+@use '../../foundations/utilities/mixins.scss' as utilities;
+
+.fudis-button-host {
```
marisanity, line 8: Is this `fudis-button-host` only used by the ngx-package for internal/logical styling? I think those styles that are only used by the angular library could be left there and add only those styles here that are needed to style the button itself

core/src/components/button/button.scss
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  &[aria-disabled='true'] {
+    @include colors.text-color('gray-dark');
+    @include colors.bg-color('gray-light');
+    @include borders.border('2px', 'solid', 'transparent');
+    @include borders.outline('1px', 'dashed', 'gray-middle');
+  }
+
+  &__label {
```
marisanity, line 38: `&--hidden` and `&--visible` styles are excatly the same? Is this class needed? Maybe padding: `0 spacing.$spacing-xs;` should be in the `.fudis-button` itself? One way to test this is just to change this first in the ngx library and run the visual tests and see if anything breaks and check it manually as well

core/src/components/button/button.stories.js
```diff
@@ -0,0 +1,51 @@
+import "./button.scss";
+
+export default {
+  title: "Components/Button",
+  decorators: [
+    (storyFn) =>
+      `<div style="display: flex; gap: 12px; flex-wrap: wrap; align-items: flex-start;">${storyFn()}</div>`,
+  ],
+};
+
```
marisanity, line 10: I think the type=button is redundant, as you don't need to declare it for the button native element right?

marisanity (approved):

PR #585, opened 2025-05-05
[Chore]: Cleaning Storybook documentation

MayaMarjut:
https://funidata.atlassian.net/browse/DS-477

Hide private Tab Navigation methods from Storybook documentation. 
Also checked various other component's controls tab and ArgTables. Ensured that provided controls were meaningful.

### DIFF
--- a/ngx-fudis/projects/ngx-fudis/src/lib/utilities/storybook.ts
+++ b/ngx-fudis/projects/ngx-fudis/src/lib/utilities/storybook.ts
@@ -335,13 +335,7 @@ export const linkExclude: RegExp = excludeRegex([
 /**
  * Notification
  */
-export const notificationExclude: RegExp = excludeRegex([
-  'articleElement',
-  'focus',
-  'link',
-  'linkTitle',
-  'externalLink',
-]);
+export const notificationExclude: RegExp = excludeRegex(['articleElement', 'focus']);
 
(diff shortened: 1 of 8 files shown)


ngx-fudis/projects/ngx-fudis/src/lib/utilities/storybook.ts
```diff
@@ -439,23 +442,30 @@ export const spacingExclude: RegExp = excludeRegex([
...
 
 /**
  * Text Input
  */
 export const textInputExclude: RegExp = excludeRegex([...formCommonDocsExclude]);
 
-export const textInputControlsExclude: RegExp = excludeRegex([...formCommonControlsExclude]);
+export const textInputControlsExclude: RegExp = excludeRegex([
+  ...formCommonControlsExclude,
+  'name',
```
RiinaKuu, line 467: Text-input `name` property actually should be visible since it is a required (or highly advised) input when using the `autocomplete` input (I think it is a good idea to enable this feature in the controls as well). Previously it wasn't working due to not existing in the story arguments.

Best to initialize `name` with empty string

```
const commonArgs: Partial<TextInputComponent> = {
...
  autocomplete: 'off',
  name: '',
};
```

and add it to the template

```
  [autocomplete]="autocomplete"
  [name]="name"
```

RiinaKuu (changes requested):

PR #591, opened 2025-05-13
[DescriptionList]: Hide compact list commas from DOM

MayaMarjut:
https://funidata.atlassian.net/browse/DS-363

Added logic to Description List Item, to determine the id of a last currently visible Description List Item Detail children and pass it to all children.

If Description List variant is compact and Description List Item Detail is last child, hide last comma from DOM.

Deleted description-list-item-details-theme.scss file, since hiding and showing commas is handled elsewhere. Also added new method 'getDlGrandChildrensIds' to id.service to return all current Description List Item Detail children.


marisanity (changes requested):
This seems like a bit complicated solution to me. I think the required logic could be just implemented using the `@ContentChildren` query in the `description-list-item.component.ts` file? Like so:

```
  @ContentChildren(DescriptionListItemDetailsComponent, {
    descendants: true,
    read: ElementRef,
  })
  ddElements: QueryList<ElementRef>;
```

In the component init, you could access this in the `AfterContentInit`-lifecycle and also use the effect in the component if the variant changes later. 

This way, you could still delete the scss-file and the logic from the `DescriptionListItemDetailsComponent`-component and would not need to add any functionality outside the `DescriptionListItem` component (like to the id-service). Of course this needs classname usage from the fudis-dd component, but with this complex component there is no pretty way to achieve this (of course there is always a full refactor but that is not an option currently...). But at least simpler way is possible! 💪 

Of course I didn't go through all the use cases, so there might be some flaws in this idea. But I made a quick poc and it seemed to work just fine. But check it out!

marisanity (approved):

PR #595, opened 2025-05-16
[CheckboxGroup, RadioButtonGroup]:Screenreader error messages

MayaMarjut:
https://funidata.atlassian.net/browse/DS-484

Binded Guidance error text content to first input of CheckboxGroup and RadioButtonGroup components. Note! Guidance helptext is now hidden from assistive technologies, but added to be a part of a fieldset legend. This was done earlier in another ticket.

Did some investigation with different screenreader browser combination, and it seems there is a difference on how chrome and firefox treat content that has been referenced elsehere with aria-described-by and wrapped within aria-hidden element. Technically this works in some cases, but doesn't seem as robust solution. Therefore decided NOT to hide whole guidance with aria-hidden true, which leads to the fact that screen reader will get error notification 2 times. ( when focusing on the first invalid input and when focusing on visual error message.)


RiinaKuu (approved):
LGTM!
I think this solution works great with the error message binded to the first group element. Also noticed some variation between browsers (Chrome, Safari, Firefox) but all of them communicated the messages to the screen reader. I didn't find it distracting that the error message is also present when focusing on the visual message.

PR #605, opened 2025-06-19
[Dependencies]: Add Core as dependency

RiinaKuu:
https://funidata.atlassian.net/browse/DS-470

- Updated `package-lock.json` files since they were not up-to-date after latest Angular update (mishap during the update)
- Added Core as dependency and peerDependency
  - Note that we want to ensure the usage of specific Core version, hence no caret (^) in the package version
- Removed all foundation style imports from `_index.scss`
    - These styles will come from Core
    - Fonts and Angular Material related theme styles are still necessary to import through Fudis itself
    - Import Core styles for Storybook


RiinaKuu:
VR tests fail because of this style block in `guidance.scss` coming from Core (which is not necessary in Fudis)

> 
> `
>   &:not(:has(&__errors)):has(&__help-text) {
>     margin-top: spacing.$spacing-xs;
>   }
> `

MayaMarjut (approved):

PR #620, opened 2025-07-11
[IdService]: Identical option label

RiinaKuu:
https://funidata.atlassian.net/browse/DS-547

Fudis is generating unique ids based on option labels. We do not recommend identical labels on multiple option but as this occured in consuming application, had to fix it.

It would have been quite heavy refactoring if I started modifying the generation logic. And if I remember correct, the id generation was first made based on the option value but since it can be string or even an object it would've been too complex --> need to stick with labels.

- Added new `labelKey` input to SelectOptionBaseDirective
- Added it as an optional parameter for `getNewSelectOptionId`  and `createSelectOptionId` functions in IdService
- Wrote documentation about how and when to use it
  - Consuming application is responsible to add the input if needed
  - If identical labels are used and `labelKey` has not been implemented, it will cause id collision and incorrect behvaviour of (at least) MultiselectComponent
    - SelectComponent seemed to handle same label selections correctly even without `labelKey` but for the sake of coherence did not separate the two and the instructions apply to both

This is easily repeatable:
In `mock_data.ts` add option with already existing label e.g.
```
  {
    value: 'value-5-armadillo_(PARTLY_ENDANGERED)_second',
    label: 'Screaming hairy armadillo (partly endangered)',
    sound: "Rollin' rollin' rollin'!",
  },

```

In `multiselect.stories.ts` add documented step (using index as `labelKey`) e.g. in ExampleTemplate


marisanity:
This solution sadly not fixing the issue. In SelectControlValueAccessor we cannot determine how to generate the id for the select option, when sorting the multiselect chips. I maybe got an idea how to fix this, as the issue is only in the behavior of PointerEvent. Of course the right choice would be to give full control of id's to the consuming application and getting rid of the idservice, but that needs a lot of time and work from us and the consuming application developers as the library is relying on it heavily. I still think it should be considered at some point though. I'll close this and make another pull request.

PR #624, opened 2025-07-17
[Components]: Circular dependencies detected in few components

RiinaKuu:
https://funidata.atlassian.net/browse/DS-508

Files affected by circular dependency:
- `button.component.ts` > `dropdown-menu.component.ts`
  - Created shared `dropdown-event.service.ts`
  - Dropdown element's width calculation has to be triggered by click coming from ButtonComponent
    - DropdownMenu subscribes to the triggering observable
  - DropdownMenu can again use parent ButtonComponent through `@Host()` decorator as before (without forwardRef injection that was earlier made as a hotfix)
- `breakpoints.ts` > `spacing.ts`
  - These were type-only imports causing circularity
  - Moved `FudisBreakpointKey` type to a separate file `breakpoint-keys.ts`
- `multiselect.component.ts` > `select-base.directive.ts` & `select-base.directive.ts` > `select.component.ts`
  - SelectBase needs to listen to some component changes from Select and Multiselect
    - Created shared interface `base-selectable.interface.ts` with generic typing and necessary properties for SelectBase to be able to use it without importing Select and Multiselect
    - Added the interface implementation to Select and Multiselect

### DIFF
--- a/ngx-fudis/projects/ngx-fudis/src/lib/components/button/button.component.ts
+++ b/ngx-fudis/projects/ngx-fudis/src/lib/components/button/button.component.ts
@@ -22,7 +22,7 @@ import {
   FudisComponentChanges,
 } from '../../types/miscellaneous';
 import { BehaviorSubject } from 'rxjs';
-import { DropdownMenuComponent } from '../dropdown-menu/dropdown-menu.component';
+import { DropdownEventService } from '../../services/dropdown/dropdown-event.service';
 
 @Component({
   selector: 'fudis-button',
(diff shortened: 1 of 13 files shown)


RiinaKuu:
Okay, it warned and skipped two files since we had one accidental `public-api` import in one of our test files. After fixing it, madge warns and skips one file and I can't figure out why. But this shouldn't be a major issue or concern since all the circularities are fixed.

```
madge --circular --extensions ts  ngx-fudis/projects/ngx-fudis/src/ --warning            
Processed 319 files (1.1s) (1 warning)

✔ No circular dependency found!

✖ Skipped 1 file

projects/ngx-fudis/src/lib/types/forms
```

PR #636, opened 2025-08-20
[Debendabot]: Configure debendbot yaml

MayaMarjut:
https://funidata.atlassian.net/browse/DS-555

Removed redundant docker check from root. Ignored major updates also from Docker-files.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -5,19 +5,11 @@
 version: 2
 updates:
   - package-ecosystem: npm
-    directory: "/"
-    schedule:
-      interval: weekly
-    groups:
-      linting:
-        patterns:
-          - "prettier"
-    ignore:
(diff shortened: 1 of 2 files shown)


.github/dependabot.yaml
```diff
@@ -79,17 +79,18 @@ updates:
...
+      - dependency-name: "*"
+        update-types: ["version-update:semver-major"]
 
   - package-ecosystem: docker
     directory: "/test"
     schedule:
       interval: weekly
+    ignore:
+      - dependency-name: "*"
+        update-types: ["version-update:semver-major"]
```
joonashak, line 96: Tautology can be avoided here by writing

```yaml
directories:
  - "/ngx-fudis"
  - "/test"
```

since the configuration is identical

joonashak (approved):

PR #651, opened 2025-08-22
Run VR tests in parallel to reduce CI run duration

joonashak:
https://funidata.atlassian.net/browse/DS-523

This PR modifies the Playwright test setup so that the tests are run in parallel in CI. This does not affect running tests locally. I use GitHub's matrix strategy and Playwrights sharding feature to split the tests into four groups and run them on different workflow runners. After the shards have completed, I combine the shard-specific blob reports into a single report.

> [!NOTE]
> Branch protection settings must be updated to require the new tests before a PR is merged. (The test names change and there is now one test per shard.)

> [!TIP]
> The blob reports are JSON and not nicely readable. When a shard fails, it's easier to wait until the combined report has been generated.

### DIFF
--- a/.github/workflows/run_tests.yaml
+++ b/.github/workflows/run_tests.yaml
@@ -1,5 +1,8 @@
 name: Run tests
 
+permissions:
+  contents: read
+
 on:
   # Enable reusing this workflow.
   workflow_call:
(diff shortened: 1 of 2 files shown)


marisanity (dismissed):
Very nice! 👍

marisanity (approved):

marisanity:
I wonder if this is yet another flaky test that fails. Will rerun and fix it in a seperate pull request if that is the case (which it probably is...)

PR #675, opened 2025-09-08
[Breadcrumbs]: Remove styles which comes from Core

RiinaKuu:
**NOTE: Merge only after new Core version is installed to Ngx** (tests should go through after that).

- Removed styles that now comes from Core
   - Changes are in this PR: https://github.com/funidata/fudis-core/pull/59


MayaMarjut (dismissed):

RiinaKuu:
Note that this PR waits newest changes from Core and therefore cannot be merged before new Core version is installed to Ngx.

krisrelax (approved):

PR #697, opened 2025-09-25
[Select]: Unable to determine if aria-controls referenced ID exists on the page while using aria-haspopup

RiinaKuu:
https://funidata.atlassian.net/browse/DS-461

Storybook accessibility checker threw an error: `Unable to determine if aria-controls referenced ID exists on the page while using aria-haspopup: aria-controls="fudis-select-1-dropdown"` even though browser's **axeDevTool did not give any errors**. 

There has been some debate whether it is necessary to add the conditionality to `aria-controls` but since our dropdown is not only hidden but it's not present in the DOM at all before user opens it the first time (after that only the visibility changes but it stays in the DOM), I think it is reasonable to add `aria-controls` only after the first opening.

-------
Some snippets from different sources:

> When a combobox is expanded, authors MUST ensure it contains or owns an element that has a role of listbox, [tree](https://www.w3.org/TR/wai-aria-1.1/#tree), grid, or [dialog](https://www.w3.org/TR/wai-aria-1.1/#dialog). This element is the combobox popup. When the combobox is expanded, authors MUST set [aria-controls](https://www.w3.org/TR/wai-aria-1.1/#aria-controls) on the textbox element to a value that refers to the combobox popup element.

https://www.w3.org/TR/wai-aria-1.1/#aria-controls

> A [combobox](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/combobox_role) element has aria-controls set to a value that refers to the element that serves as the popup. The aria-controls only needs to be set when the popup is visible, but it is valid and easier to program to reference an element that is not visible.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-controls


MayaMarjut:
Add this to multiselect too

MayaMarjut (approved):

PR #702, opened 2025-10-03
[Pagination]: New component

MayaMarjut:
https://funidata.atlassian.net/browse/DS-176

New Pagination component. Including Unit tests, VR tests and Storybook documentation.

Component is currently missing handling focus on situations when previous and next buttons are hidden.

### DIFF
--- a/ngx-fudis/projects/ngx-fudis/src/lib/utilities/storybook.ts
+++ b/ngx-fudis/projects/ngx-fudis/src/lib/utilities/storybook.ts
@@ -333,6 +333,7 @@ export const notificationArgsTableExclude: RegExp = excludeRegex(['articleElemen
 export const paginationCommonExclude: string[] = [
   'itemList',
   'hasUserChangedPage',
+  'goToPage',
   'observer',
   'prevPageIndex',
   'range',
@@ -348,7 +349,6 @@ export const paginationControlsExclude: RegExp = excludeRegex([
   ...paginationCommonExclude,
   'id',
   'pageChange',
-  'goToPage',
 ]);
 
 /**


ngx-fudis/projects/ngx-fudis/src/lib/components/pagination/pagination.mdx
```diff
@@ -0,0 +1,46 @@
...
+  [pageHref]="(i) => '/products?page=' + (i + 1)"
+  [paginationAriaLabel]="'Example navigation'"
+  (pageChange)="onPageChange($event)">
+</fudis-pagination>
+
+```
+
+### Properties
+
+Pagination requires `pageCount` in order to generate total amount of pages. Additionally current `pageIndex` can be provided, the first page starts with index 0. Pagination has also a mandatory
```
RiinaKuu, line 32: I would like to get a bit more specific documentation about `pageIndex`. I think it should clearly state what it does and as I understood it initializes the component with the given index and sets it as a starting page.

I would also organize the paragraph so that the mandatory properties come first and then the optional ones.

ngx-fudis/projects/ngx-fudis/src/lib/utilities/storybook.ts
```diff
@@ -71,14 +71,12 @@ const buttonCommonExclude: string[] = [
...
   'dropdownOpen',
   'buttonEl',
-  'handleBlur',
   'buttonClick',
   'closeMenu',
   'toggleMenu',
   'popoverTriggerLabel',
   'popoverPosition',
   'popoverText',
-  'id',
```
RiinaKuu, line 81: I think your intention here has been to add `handleBlur` and `id` to the Button Properties list, which is correct. But now these properties have been added to controls view as well. So you still need to add them to `buttonControlsExclude` and `buttonIconOnlyExclude` to exclude them from the controls view.

ngx-fudis/projects/ngx-fudis/src/lib/components/pagination/pagination.component.ts
```diff
@@ -0,0 +1,342 @@
...
+    siblingCount: number,
+  ): (Ellipsis | number)[] {
+    const startPages = this.range(1, Math.min(1, pageCount));
+    const endPages = this.range(Math.max(pageCount, 2), pageCount);
+
+    /**
+     * Calculate the first sibling page number to display around the current page
+     */
+    const siblingsStart = Math.max(
+      Math.min(pageIndex + 1 - siblingCount, pageCount - 1 - siblingCount * 2 - 1),
```
RiinaKuu, line 270: This might be personal but this is **very** hard to read and I'm not sure (without testing) does the calculation work "properly" without parentheses.

Is the calculation pattern `(pageIndex + 1) - siblingCount, ((pageCount - 1) - (siblingCount * 2)) - 1 `or something else? So could this and the rest of the similar calculations have bit more structure with the parentheses for example.

ngx-fudis/projects/ngx-fudis/src/lib/components/pagination/pagination.component.ts
```diff
@@ -0,0 +1,342 @@
...
+  /**
+   * To observe container size changes and adjust Pagination sibling count based on viewport changes
+   */
+  private observer?: ResizeObserver;
+
+  /**
+   * Computed list of pagination items (page numbers + ellipses)\
+   * Recalculated whenever page count, current page, or sibling count changes.
+   */
+  itemList = computed(() =>
```
RiinaKuu, line 227: Our first computed signal in the whole library, yey :D But should/could this also be set as `protected` like any other variables?

RiinaKuu (changes requested):
Overall looking good!
This is the first set of reviewing, I didn't go through unit test, html or styles yet but wanted to expose these comments already.

PR #728, opened 2025-10-29
[Popover]: Attach portal to the current viewContainerRef

RiinaKuu:
https://funidata.atlassian.net/browse/DS-600

Noticed that in consuming application Popover content was not accessible by screen readers. It was due to NgbModal behavior in setting content outside of modal `aria-hidden="true"`.

This fix should move the popover content from global overlay container to where the host button exists.

Tested in Sisu by switching locally to `git checkout tags/v7.0.0` and using yarn symlink to test the local changes. With this fix Popover content was accessible for screen readers inside NgbModals too.


RiinaKuu:
This fix falsely seemed to be working after clearing angular cache and testing only one NgbModal in the consuming application side. But after navigating to different views it stopped working due to overlay rendering logic (can't remember the exact reason anymore).

PR #900, opened 2026-04-16
[Expandable]: Convert to OnPush

RiinaKuu:
https://funidata.atlassian.net/browse/DS-374

Converting Expandable component to OnPush
- Updated internal `_closed` and `_openedOnce` to signals
- Added explicit lifecycle for subscription with `DestroyRef`
- Added guard against late resolve of the Promise
  
### Developer's checklist

### DIFF
--- a/ngx-fudis/projects/ngx-fudis/src/lib/components/expandable/expandable.component.html
+++ b/ngx-fudis/projects/ngx-fudis/src/lib/components/expandable/expandable.component.html
@@ -2,7 +2,7 @@
   [id]="_id"
   [attr.aria-labelledby]="_headingId"
   class="fudis-expandable fudis-expandable__{{ variant }}"
-  [class.fudis-expandable--closed]="_closed"
+  [class.fudis-expandable--closed]="_closed()"
 >
   <div class="fudis-expandable__header fudis-expandable__header__{{ variant }}">
     <div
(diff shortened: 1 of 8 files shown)


ngx-fudis/projects/ngx-fudis/src/lib/components/expandable/expandable.component.html
```diff
@@ -17,17 +17,17 @@
         class="fudis-expandable__header__heading__button fudis-expandable__header__heading__button__{{
           variant
         }}"
-        [attr.aria-expanded]="!_closed"
+        [attr.aria-expanded]="!_closed()"
         [attr.aria-controls]="!closed ? _id + '__content' : null"
```
mfaarni: Should this be `!_closed()` instead of `!_closed` as well?

RiinaKuu: This was quite confusing for me too since the naming is very similar, but this specific `closed` is referring to the `@Input` which user can set, no the internal signal.

RiinaKuu: After further discussion, decided to change the aria-controls to listen to the signal as well, since the given `@Input` is always updating the internal `_closed` signal.

PR #902, opened 2026-04-20
[RadioButton, RadioButtonGroup]: Convert to OnPush

RiinaKuu:
https://funidata.atlassian.net/browse/DS-383

Converting RadioButton and RadioButtonGroup components to OnPush
- Created and populated private signals with current FormControl states and value
- Created and exposed public readonly signals for the use of child RadioButton
- Changed `control.valueChanges` to `control.events`
  - https://angular.dev/guide/forms/reactive-forms#unified-control-state-change-events
  - From Angular's `index.d.ts` JSdoc about `events` Observable: 
    -  _A multicasting observable that emits an event every time the state of the control changes. It emits for value, status, pristine or touched changes._
  
### Developer's checklist

### DIFF
--- a/ngx-fudis/projects/ngx-fudis/src/lib/components/form/radio-button-group/radio-button-group.component.ts
+++ b/ngx-fudis/projects/ngx-fudis/src/lib/components/form/radio-button-group/radio-button-group.component.ts
@@ -11,6 +11,7 @@ import {
   DOCUMENT,
   ChangeDetectionStrategy,
   signal,
+  WritableSignal,
   Signal,
 } from '@angular/core';
 import { FudisInputSize, FudisRadioButtonChangeEvent } from '../../../types/forms';
@@ -84,10 +85,10 @@ export class RadioButtonGroupComponent
   /**
    * Private signals for managing the state and selected value of the radio button group
    */
-  private _selectedValue = signal<string | boolean | object | null | unknown>(null);
-  private _touched = signal(false);
-  private _invalid = signal(false);
-  private _disabled = signal(false);
+  private _selectedValue: WritableSignal<string | boolean | object | null | unknown> = signal(null);
+  private _touched: WritableSignal<boolean> = signal(false);
+  private _invalid: WritableSignal<boolean> = signal(false);
+  private _disabled: WritableSignal<boolean> = signal(false);
 
   /**
    * Publicly exposed readonly signals for template binding and external use (in single radio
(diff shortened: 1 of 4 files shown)


ngx-fudis/projects/ngx-fudis/src/lib/components/form/radio-button-group/radio-button-group.component.ts
```diff
@@ -78,20 +81,44 @@ export class RadioButtonGroupComponent
    */
   private _subscription: Subscription;
 
+  /**
+   * Private signals for managing the state and selected value of the radio button group
+   */
+  private _selectedValue = signal<string | boolean | object | null | unknown>(null);
+  private _touched = signal(false);
+  private _invalid = signal(false);
+  private _disabled = signal(false);
```
mfaarni: In expandable, similar signals were defined using `WritableSignal<boolean>`. Do we want to add the type declarations to all signals here too?

ngx-fudis/projects/ngx-fudis/src/lib/components/form/radio-button-group/radio-button/radio-button.component.html
mfaarni, line 5: should this label disabling control also check ` _parentGroup.disabledState() `?

ngx-fudis/projects/ngx-fudis/src/lib/components/form/radio-button-group/radio-button-group.component.ts
RiinaKuu: Yes good point, even though `signal()` is writable by default, let's be consistent and use it with the type declaration.

ngx-fudis/projects/ngx-fudis/src/lib/components/form/radio-button-group/radio-button/radio-button.component.html
RiinaKuu: This class (`.fudis-radio-button--disabled`) is only used for disabling pointer events if the disabling comes from the `@Input` (which we do not recommend, FormControl disabling is the correct way). Also found this comment from `radio-button.scss` so this disabled check does not rely on parent group control state.


> If RadioButtonGroup is disabled through "disabled" Input, we need to set pointer-event to none for each radio-button. This is not necessary when disabling comes through formControl.

mfaarni (approved):

PR #906, opened 2026-04-30
[Datepicker, DateRange]: Convert to OnPush

mfaarni:
DS-382

- Changed DatepickerComponent and DateRangeComponent to use OnPush
- Added _touched, _invalid, and _disabled signals to DatepickerComponent with a _syncControlState() method to keep them in sync
- Replaced control.valueChanges subscription with control.events for state syncing
- Kept a separate control.valueChanges subscription for date range crossing checks (stored in its own _dateCrossingSubscription field so it can be properly unsubscribed when the control changes)
  
### Developer's checklist


RiinaKuu (approved):
LGTM and everything seems to work as before 👍

PR #911, opened 2026-05-05
[Expandable]: Remove margin-bottom

emilyhernandez:
DS-669

Removed the bottom margin from Expandable and update visual regression tests. In addition:

- Added bottom margin back to expandable component's host elements in test components where expandable was part of a more complex view (for example forms) to keep the visuals the same
- Updated expandable snapshots

- Update snapshots for tests where the expandable margin bottom missing only affected the the snapshot height (ie. the expandable was the last element in the image)

### Developer's checklist


RiinaKuu (commented):
The changes look good 👍  However, the sandbox layout is now a bit broken, so it would be good to fix that as well.

Image

emilyhernandez:
Screenshot 2026-05-08 at 12 56 09

Good catch, I had missed sandbox entirely. It's now fixed 👍

RiinaKuu (approved):

PR #912, opened 2026-05-05
[Select, Multiselect]: Convert to OnPush

RiinaKuu:
DS-390

Converted the whole Select and Multiselect family to OnPush.

Few things to point out:
- **SelectOption** and **MultiselectOption**
  - Since the `data` input is coming from the parent, it is the parent's responsibility not to mutate the same object instance. Added a comment in Select and Multiselect components' JSDocs ("_Option data should be treated as immutable. ..._")
  - SelectOption and MultiselectOption components still read the data properties (`value`, `label`, `subLabel`, `disabled`) directly from the input object. This is fine if the parent replaces option objects if/when they change.
- **SelectDropdown**
  - There is a `computed()` signal which includes the use of signals, inputs and observables. Previously there has been an attempt to convert all of them to signals but the functionality didn't quite work. If this logic will be changed it has to be thoroughly tested in Sisu where a bug was detected in screen reader announcements, and syncing the helptext with correct number of visible options. When this bug was raised it was not possible to duplicate it in our Storybook environment.
    - The fix PR can be found here: https://github.com/funidata/fudis/pull/740
  - The signal recomputes when one of the signal dependencies change (i.e. `_resultsSignal` or `_filterTextSignal`) and during that recomputation, it reads the latest "plain property" values too, and this is why it works as is.
- **Case where two Multiselect components share the same option**
  - Since `control` is a public input, replacing it should be possible (i.e. assigning it with `new FormControl`)
    - This needed a small fix to work so that we support both ways; updating existing FormControl value and handling replaced FormControl instances correctly
    - The previous fix PR can be found here: https://github.com/funidata/fudis/pull/861
- **Using AbstractControl in SelectBase** 
  - Conscious choice since it is the common supertype for the minimal behavior the `_controlRef` signal needs to expose.

Extra:
- Ticket also suggested to convert **DropdownMenuGroup** component to OnPush
- Changed `_maxWidth` to signal since it was used in the **DropdownMenu** template as is.
  
### Developer's checklist


mfaarni (approved):
Well done!

PR #916, opened 2026-05-07
[Form]: Convert to OnPush

mfaarni:
DS-389

Converted the form component to use onpush. Small change done in the template, because after changing to onpush, the signal in the template was not read because of the ||-operator short-circuiting

### Developer's checklist

### DIFF
--- a/ngx-fudis/projects/ngx-fudis/src/lib/components/form/form/form.component.ts
+++ b/ngx-fudis/projects/ngx-fudis/src/lib/components/form/form/form.component.ts
@@ -123,7 +123,8 @@ export class FormComponent extends GridApiDirective implements OnInit, OnDestroy
   @Output() handleUpdatedErrorList = new EventEmitter<{ id: string; message: string }[] | null>();
 
   /**
-   * The _errorSummaryVisible signal tracks the current visibility of the error summary for template binding
+   * The _errorSummaryVisible signal tracks the current visibility of the error summary for template
+   * binding
    */
   protected _errorSummaryVisible: WritableSignal<boolean> = signal<boolean>(false);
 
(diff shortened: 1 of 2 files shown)


ngx-fudis/projects/ngx-fudis/src/lib/components/form/form/form.component.ts
RiinaKuu: This toggling does not feel ideal and could be improved. 
We still want to preserve the public input (and set the internal signal to be consistent with the input value before the subscribe; see next comment) but in this subscribe we should set the internal signal directly from the service value. 

```
.subscribe((value) => {
      this._errorSummaryVisibleSignal.set(value);
    });
```

RiinaKuu, line 140: Set the internal signal from the public input value immediately here, after the `registerNewForm` call.

`this._errorSummaryVisibleSignal.set(this.errorSummaryVisible);`

RiinaKuu: You can also decide whether this property should be renamed to just `_errorSummaryVisible` (since we haven't used the signal suffix that much) and use the `WritableSignal` syntax we have enforced.

Also update the comment. The `errorSummaryVisible` input is not used in the template anymore so the comment is a bit outdated. Something simple is enough.

RiinaKuu (changes requested):

ngx-fudis/projects/ngx-fudis/src/lib/components/form/form/form.component.ts
mfaarni: True, this was simplified :+1:

mfaarni: This was changed to reflect the current style better. Although the signal is still in the template, the comment was updated to be more simple :+1:

PR #920, opened 2026-05-13
[Guidance]: Convert to OnPush

mfaarni:
DS-676

Change Guidance component to use onPush.

- Guidance receives a FormControl or FormGroup as an input but does not bind it to a form element, but instead reads their state directly in the template (e.g. control.touched, control.errors). Since Angular's OnPush doesnt automatically detect changes in these properties, ChangeDetectorRef.markForCheck() is called via subscriptions to control.events and formGroup.events. This ensures the template changes whenever the control or form group state changes.

- _maxLengthText is converted from a BehaviorSubject to a signal, removing the need for AsyncPipe.
  
### Developer's checklist


RiinaKuu (approved):
Looking good 👍

PR #921, opened 2026-05-18
[Multiselect, Select]: Emit selectionUpdate before value change

emilyhernandez:
DS-675

Changed the order of `this.selectionUpdate.emit()` and `this.control.patchValue()` calls when entire selection is cleared from clear-button, so that the order matches the order of individual selection removal's calls. 
  
### Developer's checklist


mfaarni (approved):
🔥

PR #925, opened 2026-05-21
[Dependencies]: Update Playwright

mfaarni:
DS-661

Updated playwright to the newest version and added allowedHosts for Storybook so that only those hosts are allowed by Storybook. Similar change was made in Core. 
  
### Developer's checklist


RiinaKuu:
Will be updated in this PR: https://github.com/funidata/fudis/pull/993

PR #926, opened 2026-05-21
[Dependencies]: Update Angular 21

RiinaKuu:
DS-651

- Updated Angular packages to latest available 21 packages
  - Automatic migration removed `"lib": ["ES2021", "dom"]`  from base `tsconfig.json`
- Updated peer dependencies to require at least Angular 21
- Updated `jest` which required also:
  - Few new package installs to dev dependencies (`jest-environment-jsdom`, `jsdom`)
  - ESLint, TS, and Jest config changes
    - ESLint: Added jest globals for test files so that every .spec does not require explicit jest imports
    - TS: Created new file for resolving the jest globals
    - Jest: Change legacy preset to CommonJS preset
- New `jest-preset-angular` version required `esbuild` and since our CI image installs dependencies from Dockerfile from our `package.json` it had to be installed as dev dependency in the project as well
- Some additional clean-ups for TS configs 
  - Removed deprecated `baseUrl` (no longer required for `paths`)
    - This change required some path fixes in sandbox (internal import paths between workspaces are not allowed anymore)
  - Removed deprecated `downlevelIteration` (no longer necessary when using `"target": "ES2022"`)
  - Added `"rootDir"`, so that Typescript does not have to infer it
  - In library config, added `"include"` array and updated `"exclude"` array, since library should not include all the same paths which are in the base `tsconfig.json` (which it extends)
- Removed Zone.js import from Storybook since it is now zoneless by default and caused console warnings
--------
**VR tests:**
- There were failures due to Chromium typing race conditions and Chromium was dropping characters during synthetic typing
  - Playwright's `keyboard.type` did not satisfy the event logic of (at least) the Select component's autocomplete variant and it had to be fixed by manually dispatching the `keydown`/`keyup` event pair
  
### Developer's checklist


mfaarni (approved):
Everything works, nice!

PR #927, opened 2026-05-22
[chore]: Update prettier

emilyhernandez:
DS-661

Updated Prettier, since v3.8.2 added support for Angular v21.2 which is what Fudis is soon transitioning to (#926 ).

### Developer's checklist


package-lock.json
```diff
@@ -609,10 +609,11 @@
       "license": "MIT"
     },
     "node_modules/prettier": {
-      "version": "3.8.1",
-      "resolved": "https://registry.npmjs.org/prettier/-/prettier-3.8.1.tgz",
-      "integrity": "sha512-UOnG6LftzbdaHZcKoPFtOcCKztrQ57WkHDeRD9t/PTQtmT0NHSeWWepj6pS0z/N7+08BHFDQVUrfmfMRcZwbMg==",
+      "version": "3.8.3",
```
RiinaKuu, line 612: We usually want to keep our `package.json` and `lock`-file versions in sync, just for consistency/readability. But since we don't need to force the use of 3.8.3, we can let it be as is.

RiinaKuu:
I noticed there was [dependabot pull](https://github.com/funidata/fudis/pull/924) (with green pipeline) with the same update, so I decided to merge that, since the current developer is absent. Will close this PR.

PR #932, opened 2026-06-01
[Chore]: Update recommended extenions

RiinaKuu:
DS-662

Same extension removals as in Core PR: https://github.com/funidata/fudis-core/pull/191

- Went through our recommended extensions in VSCode
- Removed not-so-vital and unnecessary recommendations
- These extension recommendations should be treated as carefully as library dependencies

Removed extensions and some explanations:
- `formulahendry.auto-rename-tag`
  - Could improve devEx when used but not vital for workflow.
- `naumovs.color-highlight`
  - We use mainly color variables and only few files contain hex, rgba or hsl values. Does not give much, not important extension.
- `xyc.vscode-mdx-preview`
  - This was handy, but apparently the whole extension is broken and preview gives an error. There is an issue opened over two years ago. Not working, nor crucial part of the workflow.
- `nhoizey.gremlins`
  - Old as an oak, could be risky. We can rely mostly on just VSCode rendering and Prettier to handle unwanted chars.
- `tobermory.es6-string-html`
  - Not even sure what this is for, maybe some inline comment blocks. Seems unnecessary.
  
### Developer's checklist

### DIFF
--- a/settings/.vscode/extensions.json
+++ b/settings/.vscode/extensions.json
@@ -8,12 +8,7 @@
     "dbaeumer.vscode-eslint",
     "esbenp.prettier-vscode",
     "stylelint.vscode-stylelint",
-    "formulahendry.auto-rename-tag",
-    "naumovs.color-highlight",
-    "unifiedjs.vscode-mdx",
-    "xyc.vscode-mdx-preview",
-    "nhoizey.gremlins",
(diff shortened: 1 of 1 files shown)


mfaarni (approved):
Based on my understanding, both the removals and extensions left are justifiable 👍

PR #941, opened 2026-06-03
[GithubActions]: Update publish and delete documentation deployment workflows

RiinaKuu:
DS-662

As mentioned in [this PR](https://github.com/funidata/fudis/pull/937), `bobheadxi/deployments` does not run on Node.js 24 so had to make a switch to `github-script`.

- What we wanted to achieve:
  - Environment deployment link in PR after pushing new commits to the branch
  - Deployment deactivation after branch is deleted
- Why we needed the scripts:
  - The simplest approach (with only `environment.name` and `environment.url`) was not enough since we want to provide the deployed environment link. GitHub's secret-safety check was not satisfied with environments created with branch name, as we would get this error:
    - _`Skip setting environment url as environment 'Documentation for branch DS-662-test-branch-1' may contain secret.`_
  - With scripts we can set deployment status explicitly with `createDeploymentStatus` and send `environment_url` directly for the PR.
- Wrote the scripts with the help of Copilot
- Tested them locally and everything seemed to work (but the true test comes after merging anyway)

  
### Developer's checklist


.github/workflows/delete_branch_docs.yaml
```diff
@@ -17,10 +21,33 @@ jobs:
           aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
           aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
       - name: Delete docs from S3 bucket
-        run: aws s3 rm s3://${{ secrets.AWS_S3_BUCKET_NAME }}/ngx/branch/${{ github.event.ref }} --recursive
+        run: aws s3 rm s3://${{ secrets.AWS_S3_BUCKET_NAME }}/ngx/branch/${{ env.BRANCH_REF }} --recursive
       - name: Deactivate deployment
-        uses: bobheadxi/deployments@648679e8e4915b27893bd7dbc35cb504dc915bc8
+        if: always()
```
emilyhernandez, line 26: Do we need this `if` if the value is "always"?

RiinaKuu: If for some reason the previous S3 delete fails we would still want to deactivate and clean the deployment metadata, so this `always()` should execute the step even with the previous step failing.

emilyhernandez: Ah, understood ✅

emilyhernandez (approved):

PR #959, opened 2026-07-09
[VersionSelector]: Documentation pages are broken in prod environment

RiinaKuu:
DS-681 --> https://funidata.atlassian.net/browse/DS-689

Original issue:
- In production environment doc pages failed with errors like: `TypeError: Cannot read properties of undefined (reading 'e')` and `undefined is not an object (evaluating 'version_selector_stories.d.e')` depending on browser.
- Stack traces pointed to `DocsContext.resolveModuleExport`
- For some reason the DocsContainer + Canvas was fragile in production bundle and because of its global injection in `preview.js` it broke all doc files

Fixes:
- Replaced the `<Canvas />` with `iframe` story embed
  - This is less "Storybook-like" implementation but should be production-safe when directly loading a concrete story URL
- Seemed that all the `VersionSelectorStories` imports in MDX files fell redundant
- Fixed story import in `preview.js`
  - The wrong placement of e.g. `Link` shouldn't have caused any issues but it was under Components when it actually belongs to Directives
  
### Developer's checklist


mfaarni (approved):
Works on my machine!

PR #968, opened 2026-07-23
[Scripts]: Remove --with-deps from tests' postinstall

mfaarni:
DS-681

The postinstall script for tests folder included `--with-deps`, which is used to install Playwright's browser libraries. This is a Linux specific command, as MacOS handles the dependencies differently. When running Linux, this caused the npm install command to prompt the user for their password for sudo, but the npm loading spinner obstructed the input field which made the install get stuck. 

To my knowledge, the required dependencies are usually already installed. To be sure that even first timers have them, I included a command to install the deps manually in the Getting Started-section of the docs.

  
### Developer's checklist


emilyhernandez (approved):
Seems to work well locally 👍

PR #986, opened 2026-08-07
[chore]: Refactor Storybook docs version selector config

emilyhernandez:
DS-685

These changes are in preparation to bigger version selector changes (turning the version selector component into a React component and removing the version selector wrapper).

Changes made:
- Turned `.storybook/preview.js` into a TypeScript file
- Moved version selector specific code to `./storybook/addons/version-selector` directory
- Turned the current version selector config into an addon (see `.storybook/addons/version-selector/preview.ts`) used by a local preset (`.storybook/local-preset.ts`)
  - This means that the version selector config lives in the addon preview.ts, which is pulled into the local preset, which in turn is loaded in the `.storybook/main.js` config `addons`.
- Updated ESLint config to work on the storybook TypeScript files


RiinaKuu (approved):
LGTM!

PR #987, opened 2026-08-10
[improvement]: Refactor version selector in Storybook docs

emilyhernandez:
DS-685

Refactored the old Angular VersionSelector into a React component to use it in the Storybook UI toolbar's area for a more integrated look. This change makes later Angular + Storybook update easier.


ngx-fudis/.storybook/local-preset.ts
```diff
@@ -1,5 +1,9 @@
 import { fileURLToPath } from 'node:url';
 
 export function previewAnnotations(entry = []) {
```
RiinaKuu: Is this still necessary after the version selector addon is appended to the managerEntries?

ngx-fudis/.storybook/addons/version-selector/VersionSelector.tsx
```diff
@@ -0,0 +1,109 @@
...
+import { Button, WithTooltip, TooltipLinkList } from "storybook/internal/components";
+
+interface VersionsResponse {
+  versions: string[];
+  latest: string;
+}
+
+export default function VersionSelector() {
+  const [versions, setVersions] = useState<string[]>([]);
+  const [currentVersion, setCurrentVersion] = useState<string>("next");
```
RiinaKuu, line 11: Just to clarify: is this `next` just an initial value that has to be there, before it gets overridden by a real version from the `versions.json`? So basically this should never been seen, right?

RiinaKuu (changes requested):

ngx-fudis/.storybook/addons/version-selector/VersionSelector.tsx
emilyhernandez: Yes exacty, it's overridden by the versions.json 👍

ngx-fudis/.storybook/local-preset.ts
emilyhernandez: Good catch! I'll need to test it. It might not be needed anymore.

emilyhernandez: Seems it's not needed, so I've removed it. Thank you 🎉

RiinaKuu (approved):
Looks good!

PR #993, opened 2026-08-13
[dependencies]: Update Playwright

RiinaKuu:
DS-679

- Updated Playwright to latest `1.62.1` version
- Updated bunch of base images
  - First commit is all about browser-specific rendering changes which has caused elements to move some pixels up or down.
  - **DescriptionList** base images had some changes on how browsers interpret CSS `columns` property, e.g. Chrome does not care of its existance hence the layout changes in examples with disabled grid
    - I reckon it wasn't worth to dig deeper into the matter
  - **LanguageBadgeGroup** tests were failing because some browsers were not satisfied with Playwright's normal button `click()`. Updated the click helper to call the element's native `click()` which should be more reliable for all browsers.
  - **Datepicker** did not fail on CI but failed locally on Firefox. Changed `keyboard.press` to locator-based press which should be more robust. This is advised also in [Playwright documentation](https://playwright.dev/docs/api/class-keyboard#keyboard-press).
  - **DropdownMenu** had the same issues as Datepicker but locally they failed on Chrome. Changed to use locator-based presses.
- Added `allowedHosts` to `main.js` as made in Core previously. Build output was warning about not configured hosts.

### Developer's checklist


test/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM mcr.microsoft.com/playwright:v1.58.1-jammy
+FROM mcr.microsoft.com/playwright:v1.62.1-jammy
```
emilyhernandez, line 1: Hmm, should we use the noble-flavor instead, since [noble is based on Ubuntu 24.04](https://playwright.dev/docs/docker#image-tags) and the tests are run on 24.04 in the CI pipeline 🤔 Jammy is based on 22.04.
```suggestion
FROM mcr.microsoft.com/playwright:v1.62.1-noble
```

emilyhernandez: Discussed this in Slack, since Playwright is in its own container, we don't need to concern about the different Ubuntu versions for now 👍

emilyhernandez (approved):

PR #995, opened 2026-08-18
[chore]: Migrate from Storybook angular to angular-vite

emilyhernandez:
DS-690

Changed to using Storybook's `angular-vite` addon instead of `angular` (webpack).
  - Although the `angular-vite` addon is in "preview" phase instead of stable, Storybook recommends using it with Angular v21 and onward. In my opinion, our usecase of Storybook (and especially Vite underneat) is not so complex that we should be concerned about possible API changes.
  - [Read more about Storybook for Angular with Vite.](https://storybook.js.org/docs/get-started/frameworks/angular-vite)
  - Note that although Storybook uses Vite underneath, Angular still uses Webpack for everything else outside of Storybook builds.


ngx-fudis/.storybook/preview.ts
```diff
@@ -1,5 +1,6 @@
+import 'zone.js';
```
RiinaKuu: I think this should be removed. Storybook should already be Zoneless. Our dev (sandbox) environment and Jest test environment still use Zone but other than that everything should be migrated to Zoneless.

RiinaKuu (changes requested):

RiinaKuu (approved):

PR #996, opened 2026-08-18
[dependencies]: Update Angular to latest v21 version

emilyhernandez:
DS-693

- Updated Angular packages to latest v21 versions.


emilyhernandez:
Replaced by #1000

PR #998, opened 2026-08-24
[GithubActions]: Move excluded version maintenance to release.yaml

RiinaKuu:
DS-691

- Removed the excluded version management from Version Selector logic
- Defined `EXCLUDED_VERSIONS` in `release.yaml`
- Filtered the unwanted versions from the S3 derived version list before creating the version manifest
  - All the unwanted releases should continue to exists in S3 but should be filtered out from the `versions.json`

### Developer's checklist


mfaarni (approved):
Seems to work as intended! Real test is during the next release 😎

